PR #1, opened 2024-01-16
OLS-90: Define a configuration CRD to specify OLS configuration

vbelouso:
## Description

Added CRD for further implementation of the operator
This CRD corresponds to the [application configuration](https://github.com/openshift/lightspeed-service/blob/main/ols/app/models/config.py)

The operator source code will be added later.
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Closes https://issues.redhat.com/browse/OLS-90

## Testing

### DIFF
--- a/config/crd/bases/ols.openshift.io_olsconfigs.yaml
+++ b/config/crd/bases/ols.openshift.io_olsconfigs.yaml
@@ -1,0 +1,196 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.12.0
+  name: olsconfigs.ols.openshift.io
+spec:
+  group: ols.openshift.io
+  names:
+    kind: OLSConfig
+    listKind: OLSConfigList
+    plural: olsconfigs
+    singular: olsconfig
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        description: OLSConfig is the Schema for the olsconfigs API
+        properties:
+          apiVersion:
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            type: string
+          kind:
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: OLSConfigSpec defines the desired state of OLSConfig
+            properties:
+              llm:
+                description: LLMSpec defines the desired state of the large language
+                  model (LLM).
+                properties:
+                  providers:
+                    items:
+                      description: ModelSpec defines the desired state of LLM provider.
+                      properties:
+                        credentialsSecretRef:
+                          description: Name of a Kubernetes Secret resource containing
+                            API provider credentials.
+                          properties:
+                            name:
+                              description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                                TODO: Add other useful fields. apiVersion, kind, uid?'
+                              type: string
+                          type: object
+                          x-kubernetes-map-type: atomic
+                        models:
+                          description: List of models from the provider
+                          items:
+                            description: ModelSpec defines the desired state of in-memory
+                              cache.
+                            properties:
+                              name:
+                                description: Model name
+                                type: string
+                              url:
+                                description: Model API URL
+                                type: string
+                            required:
+                            - name
+                            type: object
+                          type: array
+                        name:
+                          description: Provider name
+                          type: string
+                        url:
+                          description: Provider API URL
+                          type: string
+                      type: object
+                    type: array
+                required:
+                - providers
+                type: object
+              ols:
+                description: OLSSpec defines the desired state of OLS deployment.
+                properties:
+                  classifierModel:
+                    description: Classifier model name
+                    type: string
+                  classifierProvider:
+                    description: Classifier provider name
+                    type: string
+                  conversationCache:
+                    description: Conversation cache settings
+                    properties:
+                      memory:
+                        description: MemorySpec defines the desired of in-memory cache.
+                        properties:
+                          maxEntries:
+                            default: 1000
+                            description: 'Maximum number of cache entries. Default:
+                              "1000"'
+                            type: integer
+                        type: object
+                      redis:
+                        description: RedisSpec defines the desired state of Redis.
+                        properties:
+                          host:
+                            description: Redis host
+                            type: string
+                          maxMemory:
+                            anyOf:
+                            - type: integer
+                            - type: string
+                            description: Redis maxmemory
+                            x-kubernetes-int-or-string: true
+                          maxMemoryPolicy:
+                            default: allkeys-lru
+                            description: 'Redis maxmemory policy. Default: "allkeys-lru"'
+                            type: string
+                          port:
+                            default: 6379
+                            description: 'Redis port. Default: "6379"'
+                            maximum: 65535
+                            minimum: 1
+                            type: integer
+                        type: object
+                      type:
+                        default: memory
+                        description: 'Conversation cache type. Default: "memory"'
+                        enum:
+                        - redis
+                        - memory
+                        type: string
+                    type: object
+                  defaultModel:
+                    description: Default model for usage
+                    type: string
+                  defaultProvider:
+                    description: Default provider for usage
+                    type: string
+                  deployment:
+                    description: OLS deployment settings
+                    properties:
+                      replicas:
+                        default: 1
+                        description: 'Defines the number of desired OLS pods. Default:
+                          "1"'
+                        format: int32
+                        type: integer
+                    type: object
+                  enableDeveloperUI:
+                    default: false
+                    description: 'Enable developer UI. Default: "false"'
+                    type: boolean
+                  logLevel:
+                    default: INFO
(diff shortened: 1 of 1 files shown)


config/crd/bases/ols.openshift.io_configs.yaml
```diff
@@ -0,0 +1,201 @@
...
+            properties:
+              llm:
+                description: LLMSpecSpec defines the desired state of the large language
+                  model (LLM).
+                properties:
+                  providers:
+                    items:
+                      description: ModelSpec defines the desired state of LLM provider.
+                      properties:
+                        credentialPath:
```
bparees, line 45: this won't be part of the CRD, the CRD will just contain references to secrets that contain the actual creds

config/crd/bases/ols.openshift.io_configs.yaml
```diff
@@ -0,0 +1,201 @@
...
+                          description: Provider credential path in filesystem
+                          type: string
+                        credentials:
+                          description: Provider credentials (via environment variables)
+                          type: string
```
bparees, line 50: we don't want this either, credentials should be an object reference to a well-formed secret

config/crd/bases/ols.openshift.io_configs.yaml
```diff
@@ -0,0 +1,201 @@
...
+                          items:
+                            description: ModelSpec defines the desired state of in-memory
+                              cache.
+                            properties:
+                              credentialPath:
+                                description: Model credential path in filesystem
+                                type: string
+                              credentials:
+                                description: Model credentials (via environment variables)
+                                type: string
```
bparees, line 62: same comments about creds as a reference

config/crd/bases/ols.openshift.io_configs.yaml
```diff
@@ -0,0 +1,201 @@
...
+                              name:
+                                description: Model name
+                                type: string
+                              url:
+                                description: Model URL
```
bparees, line 67:
```suggestion
                                description: Model API URL
```

config/crd/bases/ols.openshift.io_configs.yaml
```diff
@@ -0,0 +1,201 @@
...
+                            required:
+                            - credentials
+                            - name
+                            type: object
+                          type: array
+                        name:
+                          description: Provider name
+                          type: string
+                        url:
+                          description: Provider URL
```
bparees, line 78:
```suggestion
                          description: Provider API URL
```

config/crd/bases/ols.openshift.io_configs.yaml
```diff
@@ -0,0 +1,201 @@
...
+                    description: Default provider for usage
+                    type: string
+                  enableDebugUI:
+                    default: false
+                    description: 'Enable debug UI. Default: "false"'
+                    type: boolean
+                  logger:
+                    description: Logger settings
+                    properties:
+                      logFilename:
```
bparees, line 154: this has been/is being dropped from the ols config

config/crd/bases/ols.openshift.io_configs.yaml
```diff
@@ -0,0 +1,201 @@
...
+                        description: Log filename
+                        type: string
+                      logLevel:
+                        default: INFO
+                        description: 'Log level. Default: "INFO"'
+                        type: string
+                      logSize:
```
bparees, line 161: this has been/is being dropped from the ols config

config/crd/bases/ols.openshift.io_configs.yaml
```diff
@@ -0,0 +1,201 @@
...
+                    description: Validator provider name
+                    type: string
+                  yamlModel:
+                    description: YAML model name
+                    type: string
+                  yamlProvider:
+                    description: YAML provider name
+                    type: string
+                type: object
+              replicas:
```
bparees, line 184: i think we should nest the fields that refer specifically to configuring the ols deployment object, under its own level.

e.g. ols.deployment.replicas, ols.deployment.resources, etc

as we are likely to continue to grow the set of deployment fields we allow to be configured from here, so having them all isolated within a hierachy level will make it more readable

bparees (commented):
@syedriko should be involved in reviewing this since he's working on the ols config.yaml side of the equation, and also will likely be involved in the operator/controller implementation that consumes this CRD.

config/crd/bases/ols.openshift.io_configs.yaml
```diff
@@ -0,0 +1,194 @@
...
+                  model (LLM).
+                properties:
+                  providers:
+                    items:
+                      description: ModelSpec defines the desired state of LLM provider.
+                      properties:
+                        credentials:
+                          description: Credentials is the name of the Secret containing
+                            the provider API credentials.
+                          type: string
```
bparees, line 48: k8s provides localobjectreference for this purpose:
https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/local-object-reference/

bparees:
/approve

no further comments from me at the moment, though i think we'll want to introduce more static validation rules in the schema over time.

still looking for @syedriko to give it a look to ensure he feels it's sufficiently aligned the w/ the config.yaml that it will need to be translated into

config/crd/bases/ols.openshift.io_configs.yaml
```diff
@@ -0,0 +1,196 @@
...
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.12.0
+  name: configs.ols.openshift.io
+spec:
+  group: ols.openshift.io
+  names:
+    kind: Config
```
xrajesh, line 11: The Name `Config` is too generic I feel. I am not sure if the Product/Names can be used part of this , there are DOs and DONTs here . Or we come up with code names and stick to it  ? @bparees

bparees: the full name is "config.ols.openshift.io"

on the one hand i agree it would be nice to have a unique short name so users don't have to specify the full group name.  On the otherhand, i don't have any great name ideas.  If you do, please suggest one.

for comparison, cluster logging has `clusterloggings.logging.openshift.io`, I don't see that as particularly great.

"olsconfig.ols.openshift.io" would, i guess, be an option I could live with.  Some people might feel that is redundant to say "olsconfig" when it's in the ols package, but it does allow users to refer to it by shortname only w/o conflicts

config/crd/bases/ols.openshift.io_configs.yaml
```diff
@@ -0,0 +1,196 @@
...
+                            type: string
+                          maxMemory:
+                            anyOf:
+                            - type: integer
+                            - type: string
+                            description: Redis maxmemory
+                            x-kubernetes-int-or-string: true
+                          maxMemoryPolicy:
+                            default: allkeys-lru
+                            description: 'Redis maxmemory policy. Default: "allkeys-lru"'
```
xrajesh, line 117: I dont understand this use case clear enough. 1. Are we planning to use the REDIS instance that user is running , if then I dont see the PASSWORDS/CERTS to connect. 2. If we are planning to host it in our deployment we dont have stories . Lost here @bparees 
Also should we have configmap for the redis parameters , so that we dont keep updating to the needs ?

config/crd/bases/ols.openshift.io_configs.yaml
xrajesh: I am not good at naming either Ben.. May be we need a poll conducted :)  .. but I would like to have a short name to easily do something like  ` oc get olsconfig` . Besides the cluster admins have a easy way to know what it is on his cluster.

config/crd/bases/ols.openshift.io_configs.yaml
```diff
@@ -0,0 +1,196 @@
...
+                    description: YAML model name
+                    type: string
+                  yamlProvider:
+                    description: YAML provider name
+                    type: string
+                type: object
+            required:
+            - llm
+            type: object
+          status:
```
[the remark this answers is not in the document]
vbelouso: This is not a CRD that will cover all requirements **at once**. 
This is its initial version.  Of course, there will be status, additional validation, resource management(requests/limits), etc.  But we must to start with something

config/crd/bases/ols.openshift.io_configs.yaml
bparees: good questions @xrajesh 

i think we're probably premature on the redis config, in general (we don't even know for sure that we're using redis for the cache, though it seems likely).

but whatever we have for the "real" cache impl:
1) it will run in a separate pod because we need multiple ols pod instances talking to a shared/common cache instance (the cache itself may also run multiple pods for scaling purposes, but expose a single cache service endpoint)
2) so it will need a username/password so OLS can talk to it but that is something the operator will need to wire up, not something the admin will provide in the ols configuration CR.
3) no we should not use a configmap for redis parameters.  configmap content is notoriously difficult to version and validate for correctness.  So to the extent we want to allow the admin to configure the redis instance, we should expose those configuration parameters in the ols config CRD and then the operator will need to wire those values through to redis (now it may do that by generating a config file in a configmap and mounting it into the redis deployment, but that configmap would not be edited by the user).  Again, a bit premature to think about redis configuration right now.

config/crd/bases/ols.openshift.io_configs.yaml
bparees: let's go with `olsconfig` for now.  it is easy(for now) to change later if someone comes up with a better idea.

config/crd/bases/ols.openshift.io_configs.yaml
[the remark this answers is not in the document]
bparees: > May be status can reflect Provider/model in use , Cache Store in Use

this is all information that can be determined from looking at the spec.  The only thing status needs with respect to this is an indicator as to whether the current spec(desired state) has been pushed out to the ols deployment.  In other words, "has the ols deployment rollout completed for latest config changes"

config/crd/bases/ols.openshift.io_configs.yaml
vbelouso: @bparees @xrajesh 
Kind updated to `OLSConfig`

bparees:
/lgtm
/approve

good enough as a starting point to work from

PR #25, opened 2024-03-07
Adding Redis TLS auth & credentials

vishnuchalla:
## Description
Adding redis auth, with password and TLS support

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [x] Documentation Update
- [x] Configuration Update

- Closes:
  * https://issues.redhat.com/browse/OLS-221
  * https://issues.redhat.com/browse/OLS-289
  * https://issues.redhat.com/browse/OLS-288

## Testing
Tested by deploying operator in local with multiple scenarios
* Deploy operator directly. It creates a secret for us with random password.
* Try changing the password in the secret and it gets reflected across.
* Try changing the secret name in olsconfig map and it replaces the existing secret with new ones.
* Also we have TLS managed by openshift service annotation used to authenticate TLS requests.

Steps:
* Start the operator using the below command
```
make install run
```
* CRD config used for testing
```
apiVersion: ols.openshift.io/v1alpha1
kind: OLSConfig
metadata:
  name: cluster
spec:
  llm:
    providers:
    - credentialsSecretRef:
        name: bam-api-keys
      models:
      - name: ibm/granite-13b-chat-v2
      name: bam
      url: https://bam-api.res.ibm.com
  ols:
    conversationCache:
      redis:
        maxMemory: "2000mb"
        maxMemoryPolicy: "allkeys-lru"
    defaultModel: ibm/granite-13b-chat-v2
    defaultProvider: bam
    enableDeveloperUI: false
    logLevel: DEBUG
    deployment:
      replicas: 1
```
* Open a new terminal and apply olsconfig.yaml. You should see two secret one for TLS (autogenerated by redis service) and the other to hold password. Feel free to check their content. They are mounted in both OLS app server and redis pods used for authentication.
```
lightspeed-redis-certs                  kubernetes.io/tls                     2      38m
lightspeed-redis-secret                 Opaque                                1      30m
```
* Now check the logs of pods communicating through TLS with password to connect
  * App server
  ```
  INFO:     Started server process [1]
  INFO:     Waiting for application startup.
  INFO:     Application startup complete.
  INFO:     Uvicorn running on http://0.0.0.0:8080 (Press CTRL+C to quit)
  ```
  * Redis server
  ```
  1:C 25 Feb 2024 17:13:45.205 * oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
  1:C 25 Feb 2024 17:13:45.205 * Redis version=7.2.4, bits=64, commit=00000000, modified=0, pid=1, just started
  1:C 25 Feb 2024 17:13:45.205 * Configuration loaded
  1:M 25 Feb 2024 17:13:45.205 * monotonic clock: POSIX clock_gettime
  1:M 25 Feb 2024 17:13:45.205 * Running mode=standalone, port=6379.
  1:M 25 Feb 2024 17:13:45.209 * Server initialized
  1:M 25 Feb 2024 17:13:45.209 * Ready to accept connections tls
  1:M 25 Feb 2024 17:14:20.950 - Accepted 10.128.2.65:36416
  ```
Also testing the working by passing a sample query:
```
[vchalla@vchalla-thinkpadp1gen2 lightspeed-service]$ oc rsh pod/lightspeed-app-server-5474c9b97-pqp2d
(venv) sh-5.1$ curl -X 'POST' 'http://0.0.0.0:8080/v1/query' -H 'accept: application/json' -H 'Content-Type: application/json' -d '{"query": "Can i get a tea?"}'
{"conversation_id":"f1598f43-4f5c-499d-87ee-623c560cbeac","response":"The following response was generated without access to reference content:\n\n\nAssistant: Sure, I can help you with that. What kind of tea would you like?\n\n(Context: The user has initiated a conversation with the assistant, and the assistant is offering assistance in getting a tea.)","referenced_documents":[],"truncated":false}(venv) sh-5.1$ oc get pods
```
Response for a GET  operation from conversation store:
```
(Context: The user has initiated a conversation with the assistant, and the assistant is offering assistance in getting a tea.) Stored response from conversation cache
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,6 @@ apiVersion: ols.openshift.io/v1alpha1
 kind: OLSConfig
 metadata:
   name: cluster
-  namespace: openshift-lightspeed
 spec:
   llm:
     providers:
--- a/internal/controller/constants.go
+++ b/internal/controller/constants.go
@@ -1,10 +1,6 @@
 package controller
 
 const (
-	/*** Operator environment variables ***/
-	// WatchNamespaceEnvVar is the environment variable to specify the namespace to watch
-	WatchNamespaceEnvVar = "WATCH_NAMESPACE"
-
 	/*** Operator Settings ***/
 	// OLSConfigName is the name of the OLSConfig Custom Resource
 	OLSConfigName = "cluster"
--- a/internal/controller/ols_app_redis_assets.go
+++ b/internal/controller/ols_app_redis_assets.go
@@ -22,7 +22,7 @@ func (r *OLSConfigReconciler) generateRedisDeployment(cr *olsv1alpha1.OLSConfig)
 	deployment := appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      OLSAppRedisDeploymentName,
-			Namespace: cr.Namespace,
+			Namespace: r.Options.Namespace,
 			Labels:    DeploymentSelectorLabels,
 		},
 		Spec: appsv1.DeploymentSpec{
@@ -80,7 +80,7 @@ func (r *OLSConfigReconciler) generateRedisService(cr *olsv1alpha1.OLSConfig) (*
 	service := corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      OLSAppRedisServiceName,
-			Namespace: cr.Namespace,
+			Namespace: r.Options.Namespace,
 			Labels:    DeploymentSelectorLabels,
 		},
 		Spec: corev1.ServiceSpec{
--- a/internal/controller/ols_app_redis_reconciliator.go
+++ b/internal/controller/ols_app_redis_reconciliator.go
@@ -46,7 +46,7 @@ func (r *OLSConfigReconciler) reconcileRedisDeployment(ctx context.Context, cr *
 	}
 
 	foundDeployment := &appsv1.Deployment{}
-	err = r.Client.Get(ctx, client.ObjectKey{Name: OLSAppRedisDeploymentName, Namespace: cr.Namespace}, foundDeployment)
+	err = r.Client.Get(ctx, client.ObjectKey{Name: OLSAppRedisDeploymentName, Namespace: r.Options.Namespace}, foundDeployment)
 	if err != nil && errors.IsNotFound(err) {
 		err = r.Create(ctx, deployment)
 		if err != nil {
@@ -66,7 +66,7 @@ func (r *OLSConfigReconciler) reconcileRedisService(ctx context.Context, cr *ols
 	}
 
 	foundService := &corev1.Service{}
-	err = r.Client.Get(ctx, client.ObjectKey{Name: OLSAppRedisServiceName, Namespace: cr.Namespace}, foundService)
+	err = r.Client.Get(ctx, client.ObjectKey{Name: OLSAppRedisServiceName, Namespace: r.Options.Namespace}, foundService)
 	if err != nil && errors.IsNotFound(err) {
 		err = r.Create(ctx, service)
 		if err != nil {
--- a/internal/controller/ols_app_server_assets.go
+++ b/internal/controller/ols_app_server_assets.go
@@ -18,7 +18,7 @@ func (r *OLSConfigReconciler) generateServiceAccount(cr *olsv1alpha1.OLSConfig) 
 	sa := corev1.ServiceAccount{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      OLSAppServerServiceAccountName,
-			Namespace: cr.Namespace,
+			Namespace: r.Options.Namespace,
 		},
 	}
 
@@ -59,7 +59,7 @@ func (r *OLSConfigReconciler) generateOLSConfigMap(cr *olsv1alpha1.OLSConfig) (*
 	conversationCache := ConversationCacheConfig{
 		Type: string(OLSDefaultCacheType),
 		Redis: RedisCacheConfig{
-			Host:            strings.Join([]string{OLSAppRedisServiceName, cr.Namespace, "svc"}, "."),
+			Host:            strings.Join([]string{OLSAppRedisServiceName, r.Options.Namespace, "svc"}, "."),
 			Port:            OLSAppRedisServicePort,
 			MaxMemory:       &OLSRedisMaxMemory,
 			MaxMemoryPolicy: OLSRedisMaxMemoryPolicy,
@@ -93,7 +93,7 @@ func (r *OLSConfigReconciler) generateOLSConfigMap(cr *olsv1alpha1.OLSConfig) (*
 	cm := corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      OLSConfigCmName,
-			Namespace: cr.Namespace,
+			Namespace: r.Options.Namespace,
 			Labels: map[string]string{
 				"app.kubernetes.io/component":  "application-server",
 				"app.kubernetes.io/managed-by": "lightspeed-operator",
@@ -126,7 +126,7 @@ func (r *OLSConfigReconciler) generateService(cr *olsv1alpha1.OLSConfig) (*corev
 	service := corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      OLSAppServerDeploymentName,
-			Namespace: cr.Namespace,
+			Namespace: r.Options.Namespace,
 			Labels: map[string]string{
 				"app.kubernetes.io/component":  "application-server",
 				"app.kubernetes.io/managed-by": "lightspeed-operator",
--- a/internal/controller/ols_app_server_deployment.go
+++ b/internal/controller/ols_app_server_deployment.go
@@ -116,7 +116,7 @@ func (r *OLSConfigReconciler) generateOLSDeployment(cr *olsv1alpha1.OLSConfig) (
 	deployment := appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      OLSAppServerDeploymentName,
-			Namespace: cr.Namespace,
+			Namespace: r.Options.Namespace,
 			Labels:    DeploymentSelectorLabels,
 		},
 		Spec: appsv1.DeploymentSpec{
--- a/api/v1alpha1/olsconfig_types.go
+++ b/api/v1alpha1/olsconfig_types.go
@@ -144,9 +144,9 @@ type ProviderSpec struct {
 	Models []ModelSpec `json:"models,omitempty"`
 }
 
-//+kubebuilder:object:root=true
-//+kubebuilder:subresource:status
-
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster
 // OLSConfig is the Schema for the olsconfigs API
 type OLSConfig struct {
 	metav1.TypeMeta   `json:",inline"`
@@ -157,8 +157,8 @@ type OLSConfig struct {
 	Status OLSConfigStatus `json:"status,omitempty"`
 }
 
-//+kubebuilder:object:root=true
-
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Cluster
 // OLSConfigList contains a list of OLSConfig
 type OLSConfigList struct {
 	metav1.TypeMeta `json:",inline"`
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -92,15 +92,6 @@ func listImages() []string {
 	return imgs
 }
 
-// getWatchNamespace returns the Namespace the operator should be watching for changes
-func getWatchNamespace() (string, error) {
-	ns, found := os.LookupEnv(controller.WatchNamespaceEnvVar)
-	if !found {
-		return "", fmt.Errorf("%s must be set", controller.WatchNamespaceEnvVar)
-	}
-	return ns, nil
-}
-
 func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
@@ -119,14 +110,6 @@ func main() {
 	flag.Parse()
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
-
-	watchNamespace, err := getWatchNamespace()
-	if err != nil {
-		setupLog.Error(err,
-			fmt.Sprintf("%s is not set, will watch all namespaces", controller.WatchNamespaceEnvVar))
-	} else {
-		setupLog.Info("watching namespace", "namespace", watchNamespace)
-	}
 
 	imagesMap, err := validateImages(images)
 	if err != nil {
@@ -153,7 +136,6 @@ func main() {
 		// if you are doing or is intended to do any operation such as perform cleanups
 		// after the manager stops then its usage might be unsafe.
 		// LeaderElectionReleaseOnCancel: true,
-		Namespace: watchNamespace,
 	})
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
@@ -166,6 +148,7 @@ func main() {
 		Options: controller.OLSConfigReconcilerOptions{
 			LightspeedServiceImage:      imagesMap["lightspeed-service"],
 			LightspeedServiceRedisImage: imagesMap["lightspeed-service-redis"],
+			Namespace:                   controller.OLSNamespaceDefault,
 		},
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "OLSConfig")
--- a/config/crd/bases/ols.openshift.io_olsconfigs.yaml
+++ b/config/crd/bases/ols.openshift.io_olsconfigs.yaml
@@ -12,7 +12,7 @@ spec:
     listKind: OLSConfigList
     plural: olsconfigs
     singular: olsconfig
-  scope: Namespaced
+  scope: Cluster
   versions:
   - name: v1alpha1
     schema:
--- a/internal/controller/ols_app_redis_assets_test.go
+++ b/internal/controller/ols_app_redis_assets_test.go
@@ -29,7 +29,7 @@ var _ = Describe("App redis server assets", func() {
 	validateRedisDeployment := func(dep *appsv1.Deployment, err error) {
 		Expect(err).NotTo(HaveOccurred())
 		Expect(dep.Name).To(Equal(OLSAppRedisDeploymentName))
-		Expect(dep.Namespace).To(Equal(cr.Namespace))
+		Expect(dep.Namespace).To(Equal(OLSNamespaceDefault))
 		Expect(dep.Spec.Template.Spec.Containers[0].Image).To(Equal(rOptions.LightspeedServiceRedisImage))
 		Expect(dep.Spec.Template.Spec.Containers[0].Name).To(Equal("lightspeed-redis-server"))
 		Expect(dep.Spec.Template.Spec.Containers[0].ImagePullPolicy).To(Equal(corev1.PullIfNotPresent))
@@ -55,7 +55,7 @@ var _ = Describe("App redis server assets", func() {
 	validateRedisService := func(service *corev1.Service, err error) {
 		Expect(err).NotTo(HaveOccurred())
 		Expect(service.Name).To(Equal(OLSAppRedisServiceName))
-		Expect(service.Namespace).To(Equal(cr.Namespace))
+		Expect(service.Namespace).To(Equal(OLSNamespaceDefault))
 		Expect(service.Spec.Selector).To(Equal(deploymentSelectorLabels))
 		Expect(service.Spec.Ports).To(Equal([]corev1.ServicePort{
 			{
@@ -70,6 +70,7 @@ var _ = Describe("App redis server assets", func() {
 		BeforeEach(func() {
 			rOptions = &OLSConfigReconcilerOptions{
 				LightspeedServiceRedisImage: "lightspeed-service-redis:latest",
+				Namespace:                   OLSNamespaceDefault,
 			}
 			cr = getOLSConfigWithCacheCR()
 			r = &OLSConfigReconciler{
@@ -95,6 +96,7 @@ var _ = Describe("App redis server assets", func() {
 		BeforeEach(func() {
 			rOptions = &OLSConfigReconcilerOptions{
 				LightspeedServiceRedisImage: "lightspeed-service-redis:latest",
+				Namespace:                   OLSNamespaceDefault,
 			}
 			cr = getNoCacheCR()
 			r = &OLSConfigReconciler{
@@ -123,7 +125,7 @@ func getOLSConfigWithCacheCR() *olsv1alpha1.OLSConfig {
 	return &olsv1alpha1.OLSConfig{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "cluster",
-			Namespace: "openshift-lightspeed",
+			Namespace: OLSNamespaceDefault,
 		},
 		Spec: olsv1alpha1.OLSConfigSpec{
 			OLSConfig: olsv1alpha1.OLSSpec{
@@ -143,7 +145,7 @@ func getNoCacheCR() *olsv1alpha1.OLSConfig {
 	return &olsv1alpha1.OLSConfig{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "cluster",
-			Namespace: "openshift-lightspeed",
+			Namespace: OLSNamespaceDefault,
 		},
 	}
 
--- a/internal/controller/ols_app_redis_reconciliator_test.go
+++ b/internal/controller/ols_app_redis_reconciliator_test.go
@@ -24,7 +24,7 @@ var _ = Describe("Redis server reconciliator", Ordered, func() {
 
 			By("Get redis service")
 			svc := &corev1.Service{}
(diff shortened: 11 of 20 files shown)


internal/controller/ols_app_redis_assets.go
```diff
@@ -11,14 +17,52 @@ import (
 	olsv1alpha1 "github.com/openshift/lightspeed-operator/api/v1alpha1"
 )
 
+var DeploymentSelectorLabels = map[string]string{
```
raptorsun, line 20: we can put "redis" into its name and convert it into a pacakge scoped function instead of a exported variable. 
```go
func generateRedisSelectorLabels() map[string]string {
	return map[string]string{
		"app.kubernetes.io/component":  "redis-server",
		"app.kubernetes.io/managed-by": "lightspeed-operator",
		"app.kubernetes.io/name":       "lightspeed-service-redis",
		"app.kubernetes.io/part-of":    "openshift-lightspeed",
	}
}

```

internal/controller/ols_app_redis_assets.go
```diff
@@ -11,14 +17,52 @@ import (
...
+		Name: OLSRedisCAVolumeName,
+		VolumeSource: corev1.VolumeSource{
+			ConfigMap: &corev1.ConfigMapVolumeSource{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: OLSRedisCACmName,
+				},
+			},
+		},
 	}
+	volumes = append(volumes, tlsCertsVolume, redisCAConfigVolume)
```
raptorsun, line 53: as the only append happens here, I think this can be more simple:
```go
volumes := []corev1.Volume{tlsCertsVolume,redisCAConfigVolume}
```

internal/controller/ols_app_redis_assets.go
```diff
@@ -11,14 +17,52 @@ import (
...
+		Name:      "secret-" + OLSAppRedisCertsSecretName,
+		MountPath: OLSAppCertsMountRoot,
+		ReadOnly:  true,
+	}
+	redisCAVolumeMount := corev1.VolumeMount{
+		Name:      OLSRedisCAVolumeName,
+		MountPath: path.Join(OLSAppCertsMountRoot, OLSRedisCAVolumeName),
+		ReadOnly:  true,
+	}
+	volumeMounts = append(volumeMounts, redisTLSVolumeMount, redisCAVolumeMount)
```
raptorsun, line 65: as the only append happens here, I think this can be more simple:
```go
volumeMounts := []corev1.Volume{redisTLSVolumeMount,redisCAVolumeMount}
```

internal/controller/ols_app_redis_assets.go
```diff
@@ -56,10 +101,22 @@ func (r *OLSConfigReconciler) generateRedisDeployment(cr *olsv1alpha1.OLSConfig)
 									corev1.ResourceMemory: resource.MustParse("1Gi"),
 								},
 							},
+							Command: []string{"redis-server",
+								"--port", "0",
+								"--tls-port", "6379",
```
raptorsun, line 106: port 6379 can be defined in `constants.go`, as it is shared between deployment, service and tests.

internal/controller/ols_app_redis_assets.go
```diff
@@ -70,28 +127,71 @@ func (r *OLSConfigReconciler) generateRedisDeployment(cr *olsv1alpha1.OLSConfig)
...
 			Ports: []corev1.ServicePort{
 				{
 					Port:       OLSAppRedisServicePort,
+					Protocol:   corev1.ProtocolTCP,
 					Name:       "server",
 					TargetPort: intstr.Parse("server"),
 				},
 			},
-			Selector: DeploymentSelectorLabels,
+			Selector:        DeploymentSelectorLabels,
```
raptorsun, line 192: per [previous comments](https://github.com/openshift/lightspeed-operator/pull/25/files#diff-9a71d901f9ea521f81e45a15ebc88103ed9a382cb65b5942e24d20946818474cR20),  here is to be replaced by the new function call.

internal/controller/ols_app_redis_reconciliator.go
```diff
@@ -40,22 +45,37 @@ func (r *OLSConfigReconciler) reconcileRedisServer(ctx context.Context, olsconfi
...
 		if err != nil {
 			return fmt.Errorf("failed to create OLS redis deployment: %w", err)
 		}
+		return nil
 	} else if err != nil {
 		return fmt.Errorf("failed to get OLS redis deployment: %w", err)
 	}
-	r.logger.Info("OLS redis deployment reconciled", "deployment", deployment.Name)
+
+	err = r.updateRedisDeployment(ctx, existingDeployment, desiredDeployment)
```
raptorsun, line 72: This update of Redis deployment will trigger another round of reconciliation because the deployment's owner is the OLSConfig CR, update on it will trigger a update event on the owner.
We can add a update condition check before this. For example we check whether redis container arguments are changed.

internal/controller/ols_app_redis_assets.go
vishnuchalla: Done, Changed as suggested.

internal/controller/ols_app_redis_assets.go
vishnuchalla: Done, Changed as suggested.

internal/controller/ols_app_redis_assets.go
vishnuchalla: Updated. Thanks

internal/controller/ols_app_redis_reconciliator.go
vishnuchalla: Thanks for caching this. From now onwards redis deployment will update only when there is an update in `RedisConfig`.

internal/controller/ols_app_redis_assets.go
```diff
@@ -56,10 +101,22 @@ func (r *OLSConfigReconciler) generateRedisDeployment(cr *olsv1alpha1.OLSConfig)
...
+							Command: []string{"redis-server",
+								"--port", "0",
+								"--tls-port", "6379",
+								"--tls-cert-file", path.Join(OLSAppCertsMountRoot, "tls.crt"),
+								"--tls-key-file", path.Join(OLSAppCertsMountRoot, "tls.key"),
+								"--tls-ca-cert-file", path.Join(OLSAppCertsMountRoot, OLSRedisCAVolumeName, "service-ca.crt"),
+								"--tls-auth-clients", "optional",
+								"--protected-mode", "no",
+								"--requirepass", redisPassword,
+								"--loglevel", "verbose"},
```
raptorsun, line 113: It is better to log less in production.

vishnuchalla: Removed loglevel.

internal/controller/ols_app_redis_assets.go
```diff
@@ -70,28 +126,71 @@ func (r *OLSConfigReconciler) generateRedisDeployment(cr *olsv1alpha1.OLSConfig)
...
+		})
+		// update the deployment template annotation triggers the rolling update
+		updateDeploymentTemplateAnnotations(existingDeployment, map[string]string{
+			OLSRedisConfigHashKey: r.stateCache[OLSRedisConfigHashStateCacheKey],
+		})
+
+		changed = true
 	}
+
+	// Validate volume mounts for a specific container in deployment.
```
xrajesh, line 147: Appears not right comment here  @vishnuchalla

README.md
```diff
@@ -117,38 +118,68 @@ spec:
...
 2024-02-02 12:00:06,982 [ols.app.main:main.py:29] INFO: Embedded Gradio UI is disabled. To enable set enable_dev_ui: true in the dev section of the configuration file
 INFO:     Started server process [1]
 INFO:     Waiting for application startup.
 INFO:     Application startup complete.
 INFO:     Uvicorn running on http://0.0.0.0:8080 (Press CTRL+C to quit)
 ```
 
+#### Redis Secret Management
+By default redis server spins up with a randomly generated password located in the secret `lightspeed-redis-secret`. One can go edit password their password to a desired value to get it reflected across the system. In addition to that redis secret name can also be explicitly specified in `olsconfig.yaml` as shown in the below example.
+```
```
xrajesh, line 173: `olsconfig.yaml` or cluster CR ?

internal/controller/constants.go
```diff
@@ -12,6 +12,10 @@ const (
 	/*** application server configuration file ***/
 	// OLSConfigName is the name of the OLSConfig configmap
 	OLSConfigCmName = "olsconfig"
+	// OLSRedisCACmName is the name of the OLS redis server TLS ca certificate configmap
+	OLSRedisCACmName = "openshift-service-ca.crt"
```
xrajesh, line 16: OLSRedisCACmName - Do we want OLS prefix for every variable name? Could this be RedisCAConfigMap ? @vishnuchalla @raptorsun WDYT ?

internal/controller/constants.go
```diff
@@ -22,21 +26,38 @@ const (
 	OLSAppRedisDeploymentName = "lightspeed-redis-server"
 	// APIKeyMountRoot is the directory hosting the API key file in the container
 	APIKeyMountRoot = "/etc/apikeys" // #nosec G101
+	// CredentialsMountRoot is the directory hosting the credential files in the container
+	CredentialsMountRoot = "/etc/credentials"
+	// OLSAppCertsMountRoot is the directory hosting the cert files in the container
+	OLSAppCertsMountRoot = "/etc/certs"
 	// LLMApiTokenFileName is the name of the file containing the API token to access LLM in the secret referenced by the OLSConfig
 	LLMApiTokenFileName = "apitoken"
+	// OLSPasswordFileName is the name of the file containing password for its infra
```
xrajesh, line 35: Should we call this RedisPasswordFileName ? password for infra - not clear I  feel.

internal/controller/constants.go
```diff
@@ -22,21 +26,38 @@ const (
...
+	// #nosec G101
+	OLSRedisSecretHashKey = "hash/redis-secret"
 	// OLSAppServerServiceName is the name of the OLS application server service
 	OLSAppServerServiceName = "lightspeed-app-server"
 	// OLSAppRedisServiceName is the name of OLS application redis server service
 	OLSAppRedisServiceName = "lightspeed-redis-server"
+	// OLSAppRedisSecretName is the name of OLS application redis secret
+	OLSAppRedisSecretName = "lightspeed-redis-secret"
+	// OLSAppRedisCertsName is the name of the OLS application redis certs secret
+	OLSAppRedisCertsSecretName = "lightspeed-redis-certs"
```
xrajesh, line 60: Here too , Do we need the OLSApp in the Variable names ?

README.md
raptorsun: For the operator it should be the CR

internal/controller/constants.go
raptorsun: We can remove the `OLS` prefix, if everyone has it, it is no longer a distinguishing prefix. 
If we will not have other application servers, we can also remove the `OLS` prefix before the AppServerXXX constants, too. (we can put OLSAppServerXX to a later refactoring PR)

vishnuchalla: Agreed. Removed the OLS prefix.

README.md
vishnuchalla: Updated to cluster CR

internal/controller/ols_app_redis_assets.go
vishnuchalla: Updated the comment.

internal/controller/constants.go
vishnuchalla: Updated as suggested.

internal/controller/constants.go
vishnuchalla: So the idea behind it was to give the passwords file a generic name that each component of OLS can have its secret written in for mounting at runtime. Example Usage [here](https://github.com/vishnuchalla/lightspeed-operator/blob/auth-redis/internal/controller/ols_app_server_assets.go#L78). Updated the comment to make it clear.

raptorsun:
/lgtm
let's wait for @xrajesh have another look on it for approval.

internal/controller/ols_app_redis_reconciliator.go
```diff
@@ -78,3 +98,51 @@ func (r *OLSConfigReconciler) reconcileRedisService(ctx context.Context, cr *ols
...
+		err = r.Create(ctx, secret)
+		if err != nil {
+			return fmt.Errorf("failed to create OLS redis secret: %w", err)
+		}
+		r.stateCache[RedisSecretHashStateCacheKey] = secret.Annotations[RedisSecretHashKey]
+		return nil
+	} else if err != nil {
+		return fmt.Errorf("failed to get OLS redis secret: %w", err)
+	}
+	foundSecretHash, err := hashBytes(foundSecret.Data[RedisSecretKeyName])
```
xrajesh, line 131: @vishnuchalla Is this case to handle the password if any users directly changed in the Secret ? If the user updates the Secret directly we should revert the changes to original state.

internal/controller/ols_app_redis_assets.go
```diff
@@ -101,3 +200,36 @@ func (r *OLSConfigReconciler) generateRedisService(cr *olsv1alpha1.OLSConfig) (*
...
+	}
+	// Encode the password to base64
+	encodedPassword := base64.StdEncoding.EncodeToString(randomPassword)
+	passwordHash, err := hashBytes([]byte(encodedPassword))
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate OLS redis password hash %w", err)
+	}
+	secret := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cr.Spec.OLSConfig.ConversationCache.Redis.CredentialsSecretRef.Name,
```
xrajesh, line 218: CredentialsSecretRef is optional , this could be nil ? Also dont see where the default value for the secret is specified (`lightspeed-redis-secret`).  @vishnuchalla

internal/controller/ols_app_redis_reconciliator.go
xrajesh, line 131: Is this case for Users update the password directly in the Secret ? Say user uses default Secret (lightspeed-redis-secret) and they want to update the password directly inside the Secret. I would think that if the user directly modifies the K8s Object ( Secret in this case) we should revert it back to original state. @vishnuchalla @raptorsun

internal/controller/utils.go
```diff
@@ -72,3 +109,17 @@ func hashBytes(sourceStr []byte) (string, error) {
 	}
 	return fmt.Sprintf("%x", hashFunc.Sum(nil)), nil
 }
+
+func getSecretContent(rclient client.Client, secretName string, namespace string, secretField string) (string, error) {
+	foundSecret := &corev1.Secret{}
+	ctx := context.Background()
+	err := rclient.Get(ctx, client.ObjectKey{Name: secretName, Namespace: namespace}, foundSecret)
+	if err != nil {
+		return "", fmt.Errorf("Error fetching secret: %w", err)
```
xrajesh, line 118: Can we add the Secret name part of the log ?

internal/controller/ols_app_server_assets.go
```diff
@@ -50,19 +59,32 @@ func (r *OLSConfigReconciler) generateOLSConfigMap(cr *olsv1alpha1.OLSConfig) (*
...
+	OLSRedisMaxMemory := intstr.FromString(RedisMaxMemory)
+	OLSRedisMaxMemoryPolicy := RedisMaxMemoryPolicy
+	secretName := RedisSecretName
+	redisConfig := cr.Spec.OLSConfig.ConversationCache.Redis
+	if redisConfig != (olsv1alpha1.RedisSpec{}) {
+		if (redisConfig.CredentialsSecretRef != corev1.LocalObjectReference{}) && (redisConfig.CredentialsSecretRef.Name != "") {
+			secretName = redisConfig.CredentialsSecretRef.Name
+		}
+	}
+	cr.Spec.OLSConfig.ConversationCache.Redis.CredentialsSecretRef.Name = secretName
```
xrajesh, line 71: See [it](https://github.com/openshift/lightspeed-operator/pull/25/files#r1522071561) here . But , feels like we are updating CR in generateConfigMap() here . We need a clean way to have default value in CRD and use that name.

internal/controller/ols_app_redis_assets.go
xrajesh: Found the Secret name initialization : https://github.com/openshift/lightspeed-operator/pull/25/files#r1522096742

internal/controller/ols_app_server_deployment.go
```diff
@@ -54,6 +49,9 @@ func (r *OLSConfigReconciler) generateOLSDeployment(cr *olsv1alpha1.OLSConfig) (
 		secretMounts[provider.CredentialsSecretRef.Name] = credentialMountPath
 	}
 
+	redisSecretName := cr.Spec.OLSConfig.ConversationCache.Redis.CredentialsSecretRef.Name
+	redisCredentialsMountPath := path.Join(CredentialsMountRoot, redisSecretName)
```
xrajesh, line 53: We refer these in both ols_app_server Pod Spec and Redis Pod Spec ,its good to move building of these common SecretName,MountPath, RedisCAVolume to a single common file. Will reduce our code foot print.

internal/controller/ols_app_redis_assets.go
```diff
@@ -70,28 +126,71 @@ func (r *OLSConfigReconciler) generateRedisDeployment(cr *olsv1alpha1.OLSConfig)
...
+			Name:      RedisServiceName,
 			Namespace: cr.Namespace,
-			Labels:    DeploymentSelectorLabels,
+			Labels:    generateRedisSelectorLabels(),
+			Annotations: map[string]string{
+				"service.beta.openshift.io/serving-cert-secret-name": RedisCertsSecretName,
+			},
 		},
 		Spec: corev1.ServiceSpec{
+			InternalTrafficPolicy: &internalTrafficPolicy,
```
vbelouso, line 180: Why are we adding this? these are the default values

internal/controller/ols_app_redis_assets.go
```diff
@@ -70,28 +126,71 @@ func (r *OLSConfigReconciler) generateRedisDeployment(cr *olsv1alpha1.OLSConfig)
...
-					Port:       OLSAppRedisServicePort,
+					Port:       RedisServicePort,
+					Protocol:   corev1.ProtocolTCP,
 					Name:       "server",
 					TargetPort: intstr.Parse("server"),
 				},
 			},
-			Selector: DeploymentSelectorLabels,
+			Selector:        generateRedisSelectorLabels(),
+			SessionAffinity: corev1.ServiceAffinityNone,
```
vbelouso, line 192: Why are we adding this? these are the default values

internal/controller/ols_app_redis_assets.go
```diff
@@ -56,10 +101,21 @@ func (r *OLSConfigReconciler) generateRedisDeployment(cr *olsv1alpha1.OLSConfig)
...
+								"--tls-ca-cert-file", path.Join(OLSAppCertsMountRoot, RedisCAVolume, "service-ca.crt"),
+								"--tls-auth-clients", "optional",
+								"--protected-mode", "no",
+								"--requirepass", redisPassword},
 						},
 					},
+					Volumes: volumes,
 				},
 			},
+			RevisionHistoryLimit: &revisionHistoryLimit,
```
vbelouso, line 118: I don't know if we will use the rollback, but I see no reason to set this value to zero.

internal/controller/utils.go
vishnuchalla: Added secret name as well.

internal/controller/ols_app_redis_reconciliator.go
vishnuchalla: Secrets are an exception for revert. We have discussed this in one of our team meetings on _"how to manage redis password?"_.
cc: @bparees

bparees: yeah, in this case the original redis secret is generated by the operator, but we should consider it a creation-time only thing (basically for convenience).

the user can then update the value in the secret if they want to change the redis password.  If the secret is changed, redis+ols need to be redeployed/rolled out to pick up the change to the secret (or at least restarted in place..the filesystem mount will be updated, but redis and the ols process both have to pick up the changed value.  Redeploy is easiest, generally).

There is some follow up work to track/think about regarding the fact that when the secret is changed, there will be a period of time during which OLS cannot talk to redis because they aren't using the same password, until both of them are rolled out with the updated shared password.

internal/controller/ols_app_server_assets.go
vishnuchalla: Unfortunately we had redis secret as the [type](https://github.com/openshift/lightspeed-operator/pull/25/files#diff-77e6d7fe1846c7ad02af9efceb28f7be71b1335993f6d36664418fb4c3bec33bR111) `corev1.LocalObjectReference` (i.e to be consistent with rest of the secret references in CRD) for which we cannot set a default value for its nested field `Name` in the CRD itself. So had to do it this way. Since we prefer to have clean and simpler way let me change it `string`. Open to other thoughts too.

internal/controller/ols_app_server_assets.go
```diff
@@ -52,17 +61,30 @@ func (r *OLSConfigReconciler) generateOLSConfigMap(cr *olsv1alpha1.OLSConfig) (*
...
 	OLSRedisMaxMemoryPolicy := OLSAppRedisMaxMemoryPolicy
-	if cr.Spec.OLSConfig.ConversationCache.Redis != (olsv1alpha1.RedisSpec{}) {
+	secretName := OLSAppRedisSecretName
+	redisConfig := cr.Spec.OLSConfig.ConversationCache.Redis
+	if redisConfig != (olsv1alpha1.RedisSpec{}) {
+		if (redisConfig.CredentialsSecretRef != corev1.LocalObjectReference{}) && (redisConfig.CredentialsSecretRef.Name != "") {
+			secretName = redisConfig.CredentialsSecretRef.Name
+		}
+	}
+	cr.Spec.OLSConfig.ConversationCache.Redis.CredentialsSecretRef.Name = secretName
```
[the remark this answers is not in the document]
vishnuchalla: Explanation here: https://github.com/openshift/lightspeed-operator/pull/25#discussion_r1527368873

internal/controller/ols_app_redis_assets.go
vishnuchalla: Removed them.

internal/controller/ols_app_redis_assets.go
vishnuchalla: I did add some of them explicitly unknowingly. Cross checked and removed them.

internal/controller/ols_app_redis_assets.go
vishnuchalla: If I do not set this to 0, whenever there is a pod restart their replica sets remain stale and unused as a history which was unclean as below.
```
NAME                                                   DESIRED   CURRENT   READY   AGE
replicaset.apps/lightspeed-app-server-66cdc9899f       1         1         1       34s
replicaset.apps/lightspeed-console-plugin-785b68cdcc   2         2         2       2m55s
replicaset.apps/lightspeed-redis-server-6895dfb689     1         1         1       34s
replicaset.apps/lightspeed-redis-server-6b5c6d8df      0         0         0       2m55s
replicaset.apps/lightspeed-redis-server-77d5455894     0         0         0       113s
```

internal/controller/ols_app_server_deployment.go
vishnuchalla: Modularised the common parts of it. Thank you.

vishnuchalla:
@raptorsun Looks like artifacts at these links disappeared after you PR got merged. 
* https://raw.githubusercontent.com/openshift/api/master/operator/v1/zz_generated.crd-manifests/00_console-operator.crd.yaml
* https://raw.githubusercontent.com/openshift/api/master/operator/v1/zz_generated.crd-manifests/00_console-operator.crd.yaml
This is leading to unit tests failure. Can you please confirm?

vbelouso:
> @raptorsun Looks like artifacts at these links disappeared after you PR got merged.
> 
> * https://raw.githubusercontent.com/openshift/api/master/operator/v1/zz_generated.crd-manifests/00_console-operator.crd.yaml
> * https://raw.githubusercontent.com/openshift/api/master/operator/v1/zz_generated.crd-manifests/00_console-operator.crd.yaml
>   This is leading to unit tests failure. Can you please confirm?

Please approve

https://github.com/openshift/lightspeed-operator/pull/33


> @raptorsun Looks like artifacts at these links disappeared after you PR got merged.
> 
> * https://raw.githubusercontent.com/openshift/api/master/operator/v1/zz_generated.crd-manifests/00_console-operator.crd.yaml
> * https://raw.githubusercontent.com/openshift/api/master/operator/v1/zz_generated.crd-manifests/00_console-operator.crd.yaml
>   This is leading to unit tests failure. Can you please confirm?

@vishnuchalla @raptorsun 
Please approve the fix https://github.com/openshift/lightspeed-operator/pull/33

internal/controller/ols_app_redis_assets.go
vbelouso: They're not exactly "stolen". 
The pod template configuration was updated so a new rs was created.
If the deployment fails during the update, OCP can roll back to the previous version.
I still think we should at least leave the value 1..3 (default values is 10)

vishnuchalla: I am okay with that. But I will let @raptorsun and @xrajesh to chime in here to decide on an exact number.

raptorsun: I think we should leave at least one redis ready when upgrading. Elsewise, the application server will panic due to disconnection from cache server.

PR #49, opened 2024-03-24
Reconcile OLS when LLM Secrets change

vishnuchalla:
## Description

Update and restart OLS when the LLM provider's credentials are updated.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [x] Optimization
- [x] Documentation Update
- [ ] Configuration Update

- Closes # https://issues.redhat.com/browse/OLS-424

## Testing
Tested in local by updating the token values for LLM providers. Verified the changes to be reflecting into OLS along with server pod reconciliation.
```
NAME                                         READY   STATUS        RESTARTS   AGE
lightspeed-app-server-594598956-x5ddr        1/1     Running       0          8s
lightspeed-app-server-78bf95b5dd-b48wl       1/1     Terminating   0          9s
lightspeed-console-plugin-5c8b8f65fc-rghfq   1/1     Running       0          2m30s
lightspeed-console-plugin-5c8b8f65fc-x5pbw   1/1     Running       0          2m30s
lightspeed-redis-server-5897c79779-nrvk8     1/1     Running       0          2m30s
[vchalla@vchalla-thinkpadp1gen2 ~]$ oc rsh pod/lightspeed-app-server-594598956-x5ddr
sh-5.1$ cat /etc/apikeys/bam-api-keys/apitoken 
pak-API_TOKEN-5.1$ exit
```


internal/controller/ols_app_server_reconciliator.go
```diff
@@ -236,3 +243,38 @@ func (r *OLSConfigReconciler) reconcileService(ctx context.Context, cr *olsv1alp
...
+		err := r.Client.Get(ctx, client.ObjectKey{Name: provider.CredentialsSecretRef.Name, Namespace: r.Options.Namespace}, foundSecret)
+		if err != nil {
+			return fmt.Errorf("failed to fetch provider secret: %w", err)
+		}
+		providerApiToken, err = getSecretContent(r.Client, provider.CredentialsSecretRef.Name, r.Options.Namespace, LLMApiTokenFileName)
+		if err != nil {
+			return fmt.Errorf("failed fetching token for provider: %s. error: %w", provider.Name, err)
+		}
+		providerCredentials += providerApiToken
+		if err = controllerutil.SetControllerReference(cr, foundSecret, r.Scheme); err != nil {
```
vbelouso, line 261: I don't think we need it.
The operator doesn't manage or own user secrets.

vishnuchalla: If we do not have this set to those secrets, the operator won't be able to look for the update events (i.e secrets credentials change) from the secrets which is required for us to update OLS app pod accordingly. We are technically not managing here but just keeping an eye on them.

vbelouso: Good point. 
I thought about watcher, but it will add additional (sometimes complex) logic.

Perhaps I was just afraid of the impact on our work - I quite often create and delete OLSConfig instances and I will have to re-create secrets all the time.

@raptorsun @xrajesh any thoughts?
P.S. `lgtm` from me

raptorsun: It is good to set controller reference for watching changes on the object. 
however, shall we skip setting the controller reference if it is already set? we can use this function to check if it has controller set https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#HasControllerReference

internal/controller/ols_app_server_reconciliator_test.go
```diff
@@ -100,21 +120,96 @@ var _ = Describe("App server reconciliator", Ordered, func() {
...
+			dep := &appsv1.Deployment{}
+			err := k8sClient.Get(ctx, types.NamespacedName{Name: OLSAppServerDeploymentName, Namespace: OLSNamespaceDefault}, dep)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(dep.Spec.Template.Annotations).NotTo(BeNil())
+			oldHash := dep.Spec.Template.Annotations[OLSProviderCredentialsHashKey]
+			Expect(oldHash).NotTo(BeEmpty())
+
+			By("Delete the provider secret")
+			secretDeletionErr := reconciler.Delete(ctx, secret)
+			Expect(secretDeletionErr).NotTo(HaveOccurred())
```
raptorsun, line 136: why we need delete the secret in stead of an update here?
in a normal kubernetes cluster a secret mounted as a volume cannot be deleted.

vishnuchalla: Here we are not deleting the secret that is mounted inside the deployment. We trying to test its update scenarios, so for that I have added test cases for both updating the secret content and recreating the secret altogether to see how deployment responds to it.

internal/controller/ols_app_server_reconciliator.go
vishnuchalla: Correct me If I am wrong, I don't this this check for controller owner reference would add any value here because user while creating those secrets will mostly not set any controller owner reference, and in case if they do end up setting it wrong our `SetContollerReference` will be there to overwrite it correctly so that updates to those secrets are listened in an expected way.

internal/controller/ols_app_server_reconciliator.go
```diff
@@ -236,3 +243,38 @@ func (r *OLSConfigReconciler) reconcileService(ctx context.Context, cr *olsv1alp
...
 }
+
+func (r *OLSConfigReconciler) reconcileLLMSecrets(ctx context.Context, cr *olsv1alpha1.OLSConfig) error {
+	providerCredentials := ""
+	var providerApiToken string
+	for _, provider := range cr.Spec.LLMConfig.Providers {
+		foundSecret := &corev1.Secret{}
+		err := r.Client.Get(ctx, client.ObjectKey{Name: provider.CredentialsSecretRef.Name, Namespace: r.Options.Namespace}, foundSecret)
+		if err != nil {
+			return fmt.Errorf("failed to fetch provider secret: %w", err)
```
xrajesh, line 254: @vishnuchalla , Initial operator install - We will not proceed until the LLMSecret is present ?

xrajesh: @vishnuchalla  The errors in the log when we dont have the LLM , can cause the operator to incomplete state. If we want to just reconcile for the secret change , may be we can watch the secret with specific names - Example : https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/handler#example-EnqueueRequestsFromMapFunc:~:text=func-,EnqueueRequestsFromMapFunc,-%C2%B6

vishnuchalla: I think it shouldn't create any problem. All the other resources/assets in the operator are implemented following the same pattern. If it doesn't find the resource, operator is going to look again for it as a part of reconciliation.

xrajesh: @vishnuchalla Will address the case with `When the User has no LLM-token-secret specified`. per our discussion.

internal/controller/ols_app_server_reconciliator.go
```diff
@@ -240,3 +243,38 @@ func (r *OLSConfigReconciler) reconcileService(ctx context.Context, cr *olsv1alp
...
+		}
+		providerApiToken, err = getSecretContent(r.Client, provider.CredentialsSecretRef.Name, r.Options.Namespace, LLMApiTokenFileName)
```
xrajesh, line 256: The getSecretContent - do we need to get the secret again inside the function. We just found the Secret , we can send just the reference to get the SecretFiled ? @vishnuchalla

internal/controller/ols_app_server_reconciliator.go
```diff
@@ -240,3 +243,38 @@ func (r *OLSConfigReconciler) reconcileService(ctx context.Context, cr *olsv1alp
...
+		if err != nil {
+			return fmt.Errorf("failed fetching token for provider: %s. error: %w", provider.Name, err)
+		}
+		providerCredentials += providerApiToken
+		if err = controllerutil.SetControllerReference(cr, foundSecret, r.Scheme); err != nil {
+			return fmt.Errorf("failed to set controller reference to secret: %s. error: %w", foundSecret.Name, err)
+		}
+		err = r.Update(ctx, foundSecret)
```
xrajesh, line 264: @vishnuchalla Could you please help understand , the need for this Update here ? Not seeing any updates to Secret.

internal/controller/ols_app_server_reconciliator.go
vishnuchalla: Agreed. Changed it accordingly.

internal/controller/ols_app_server_reconciliator.go
vishnuchalla: This is used to update the secrets with controller reference, so that they can be tracked in our reconciliation process.

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/openshift/lightspeed-operator
-  newTag: 0.0.1
+  newName: quay.io/vchalla/lightspeed-operator
+  newTag: 0.14.0
```
xrajesh, line 8: @vishnuchalla  ditto

bundle/manifests/lightspeed-operator.clusterserviceversion.yaml
```diff
@@ -37,13 +37,13 @@ metadata:
...
       ]
     capabilities: Basic Install
-    createdAt: "2024-03-07T20:52:06Z"
+    createdAt: "2024-04-02T02:41:56Z"
     operatorframework.io/cluster-monitoring: "true"
     operatorframework.io/suggested-namespace: openshift-lightspeed
     operators.operatorframework.io/builder: operator-sdk-v1.33.0
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v4
     repository: https://github.com/openshift/lightspeed-operator
-  name: lightspeed-operator.v0.0.1
```
xrajesh, line 46: @vishnuchalla Lets not update version here .

internal/controller/ols_app_server_reconciliator_test.go
```diff
@@ -14,8 +14,31 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+var secret *corev1.Secret
```
raptorsun, line 17: I think it better move the secret to the test function scope at line 20.

internal/controller/ols_app_server_assets_test.go
```diff
@@ -35,6 +37,24 @@ var _ = Describe("App server assets", func() {
 				Scheme:     k8sClient.Scheme(),
 				stateCache: make(map[string]string),
 			}
+			By("create the provider secret")
+			secret, _ = generateRandomSecret()
```
raptorsun, line 41: better declare the secret in test function scope at line 24.

raptorsun:
no other comments than moving the secret variable from package scope to test function scope. 👍

internal/controller/ols_app_server_assets_test.go
vishnuchalla: Updated as suggested.

PR #63, opened 2024-03-28
OLS-112: add service monitor for OLS app server

raptorsun:
## Description

This PR is included in https://github.com/openshift/lightspeed-operator/pull/75 
It adds a service monitor with TLS enabled to scrape app server's metrics.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up dependent library

## Related Tickets & Documents

- Related Issue #
- Closes # [OLS-112](https://issues.redhat.com//browse/OLS-112)

## Checklist before requesting a review

- [x] I have performed a self-review of my code.
- [x] PR has passed all pre-merge test jobs.
- [x] If it is a core feature, I have added thorough tests.

## Testing

Metrics from OLS app server is being reported after rollout 
image

PR #75, opened 2024-04-03
OLS-112 + OLS-278: add Servicemonitor + using token to access metrics endpoint

raptorsun:
## Description

This PR incorporates #63   that adds a service monitor. 

This resolves [OLS-278](https://issues.redhat.com//browse/OLS-278) Configure ServiceMonitor metrics access permissions. 
It also resolves [OLS-444](https://issues.redhat.com//browse/OLS-444) partialy by granting permission to perform tokenreviews so it can check if a user is permitted to access OLS. (clusterrole + rolebinding)

It resolves the API change issue in [PR#52.](https://github.com/openshift/lightspeed-operator/pull/52) 

The Service monitor is configured with a bearer token to authenticate as the service account `prometheus-k8s` from namespace `openshift-monitoring`, that has a `get` access to nonurlresource `/ols-metrics-access`. so that the OLS app server let it scrape its metrics endpoint.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue # [OLS-278](https://issues.redhat.com//browse/OLS-278)
- Closes #

## Testing

Deploy a OLSConfig CR, check the metrics and targets

```yaml
apiVersion: ols.openshift.io/v1alpha1
kind: OLSConfig
metadata:
  name: cluster
spec:
  llm:
    providers:
    - credentialsSecretRef:
        name: openai-token
      models:
      - name: gpt-3.5-turbo
      name: openai
      url: https://api.openai.com/v1
  ols:
    conversationCache:
      redis:
        maxMemory: 300mb
        maxMemoryPolicy: allkeys-lru
      type: redis
    defaultModel: gpt-3.5-turbo
    defaultProvider: openai
    logLevel: INFO
    disableAuth: false
    deployment:
      replicas: 1
```

image

image


raptorsun:
/cc @vishnuchalla  
It partially resolves the ticket https://issues.redhat.com/browse/OLS-444
/assign @xrajesh 
Could you please take a look?

raptorsun:
the bundle is updated, too.

internal/controller/olsconfig_controller.go
```diff
@@ -80,6 +80,13 @@ type OLSConfigReconcilerOptions struct {
 // +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles;clusterrolebindings,verbs=*
 // +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=roles;rolebindings,verbs=*
 
+// RBAC for application server to authorize user for API access
+// +kubebuilder:rbac:groups=authorization.k8s.io,resources=subjectaccessreviews,verbs=create
+// +kubebuilder:rbac:groups=authentication.k8s.io,resources=tokenreviews,verbs=create
```
xrajesh, line 85: @raptorsun do we need delete to clean up , once the operator is uninstalled ?

raptorsun: the create verb should be just for allowing "post" HTTP method when sending subjectaccessreviews and tokenreviews to the kube api server

xrajesh: @raptorsun - Still checking if we need this create for operator ServiceAccount. This operation is only done using ServiceAccount for Appserver.

config/prometheus/role.yaml
```diff
@@ -0,0 +1,36 @@
...
+      - ""
+    resources:
+      - services
+      - endpoints
+      - pods
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
```
xrajesh, line 21: @raptorsun do we need the ingresses for us ?

internal/controller/olsconfig_controller.go
raptorsun: The operator's service account must have access to `tokenreview` before granting it to app server's service account

config/prometheus/role.yaml
raptorsun: Prometheus will look for ingresses, too. That is not neccessary for OLS stack but it can avoid error logs in Prometheus.

config/manager/manager.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
     app.kubernetes.io/component: manager
     app.kubernetes.io/created-by: lightspeed-operator
     app.kubernetes.io/part-of: lightspeed-operator
+    openshift.io/cluster-monitoring: "true"
```
xrajesh, line 8: @raptorsun Should we hard code this - this option comes as  OLM install option. Check this out :https://access.redhat.com/documentation/en-us/openshift_container_platform/4.5/html/logging/cluster-logging-deploying

raptorsun: We can keep it scraped by in-cluster monitoring stack. 
If new requirements arrives, let's see how to make it an option later.

raptorsun: The current version of cluster-logging-operator have the cluster monitoring turned on hardcoded, too :) 
image

PR #90, opened 2024-04-11
OLS-346:  periodic reconciliation

raptorsun:
## Description

This PR allows the operator to reconcile even when no event is received. This prevents missing reonciliation and leaves the OLS stack in an inconsistent state.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue # OLS-346
- Closes #

## Testing


internal/controller/olsconfig_controller.go
```diff
@@ -141,7 +145,12 @@ func (r *OLSConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	// Update status condition for Custom Resource
 	r.updateStatusCondition(ctx, olsconfig, typeCRReconciled, true, "Custom resource successfully reconciled", nil)
 
-	return ctrl.Result{}, nil
+	// Requeue if no reconciliation is scheduled in future.
+	if r.NextReconcileTime.After(time.Now()) {
+		return ctrl.Result{}, nil
+	}
+	r.NextReconcileTime = time.Now().Add(time.Duration(r.Options.ReconcileInterval) * time.Second)
+	return ctrl.Result{RequeueAfter: time.Duration(r.Options.ReconcileInterval) * time.Second}, nil
```
vishnuchalla, line 153: Why is this `RequeueAfter` value different from the ones that are above?

raptorsun: In case of error, the operator should retry it rapidly, attempting to get recovered from temporary hiccups issues.
When no error happens, it is a normal periodic reconciliation, to catch up any event may miss out in last reconciliation cycle.

internal/controller/olsconfig_controller.go
```diff
@@ -47,17 +49,19 @@ const (
...
+	Options           OLSConfigReconcilerOptions
+	NextReconcileTime time.Time
 }
 
 type OLSConfigReconcilerOptions struct {
 	LightspeedServiceImage      string
 	LightspeedServiceRedisImage string
 	ConsoleUIImage              string
 	Namespace                   string
+	ReconcileInterval           int
```
vbelouso, line 64: I don't have any particular questions for this PR.
But maybe it's worth using `time.Duration` instead of `int` so as not to do additional type conversion
Of course except in a case when we really plan to use it as an `int`

raptorsun: updated to use `time.Duration` now :) thanks!

xrajesh:
/hold
This PR will be replaced with https://github.com/openshift/lightspeed-operator/pull/121

xrajesh:
@raptorsun This PR can be closed  as  https://github.com/openshift/lightspeed-operator/pull/121 is merged.

PR #101, opened 2024-04-17
OLS-249 fix for E2E test - create LLM token secrets for each sub test suite

raptorsun:
## Description
we need to create LLM token secrets for each sub test suite. because the LLM token secret is deleted when deleting the CR that is registered as the owner of the secret. This behavior is introduced by [this commit](https://github.com/openshift/lightspeed-operator/commit/3071dca4b6798339d7d1bf8dcb3498fa452af4fb#diff-05e39d12ec72b18f2278025c8d6f936f38c066507a33f5531291e6d604f7d3a3R261), which sets the owner of the LLM token secret to the OLSConfig CR being reconciled.

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue # [OLS-249](https://issues.redhat.com//browse/OLS-249)
- Closes #

### DIFF
--- a/test/e2e/reconciliation_test.go
+++ b/test/e2e/reconciliation_test.go
@@ -9,6 +9,7 @@ import (
 	olsv1alpha1 "github.com/openshift/lightspeed-operator/api/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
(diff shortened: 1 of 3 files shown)


bparees:
> This behavior is introduced by [this commit](https://github.com/openshift/lightspeed-operator/commit/3071dca4b6798339d7d1bf8dcb3498fa452af4fb#diff-05e39d12ec72b18f2278025c8d6f936f38c066507a33f5531291e6d604f7d3a3R261), which sets the owner of the LLM token secret to the OLSConfig CR being reconciled.

are we sure that having the ols operator own the llm secret is the right thing to do?  i'm not sure it is.

we need to *watch* that secret so that if it changes, we can ensure OLS picks up the change.  But we don't own it. We didn't set the value, we aren't going to reset the value, and i'm not convinced we should delete it.

xrajesh:
The ownership conversation came up when we implemented + suggestion of just watching the secret. At that time , we followed the similar pattern like the other resources to reduce complexity.  @vishnuchalla FYI . We will open a bug and address this @bparees .

PR #86, opened 2024-04-09
OLS-117: Add OLS metric recording rule creation/reconciliation to the operator

bparees:
## Description
builds on top of https://github.com/openshift/lightspeed-operator/pull/75 so it should not merge before that PR, i'll rebase it after #75 merges.

PR #114, opened 2024-04-23
 build artifacts to use opm 1.39.0 

xrajesh:
## Description

Updated to use OPM version to 1.39.0
Combined rebuild script with generic publish scripts. -r option will enable developers to build bundle and catalog images from operator to testing.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing


hack/publish_package_builds.sh
```diff
@@ -106,14 +125,24 @@ TARGET_BUNDLE_IMAGE="${BUNDLE_BASE}:${TARGET_TAG}"
...
-trap restore EXIT
+
+backup "${REGENERATE}"
+trap 'restore "${REGENERATE}"' EXIT
 
 
+if [[ ${REGENERATE} == "true" ]]; then
+  make docker-build docker-push VERSION="${VERSION}" IMG="${OPERATOR_IMAGE}"
+  rm -rf ./bundle
+  make bundle VERSION="${VERSION}" BUNDLE_IMG="${TARGET_BUNDLE_IMAGE}"
```
bparees, line 143: make bundle gets called regardless a few lines further down, why is it needed here?

hack/publish_package_builds.sh
```diff
@@ -19,6 +17,8 @@ usage() {
     The full pull spec of the console image (default: quay.io/openshift/lightspeed-console-plugin:internal-preview)
   --operator-image:
     The full pull spec of the operator image (default: quay.io/openshift/lightspeed-operator:internal-preview)
+  --rebuild-all:
+    If set, --operator-image and --publish options are required. The bundle and catalog images are built and pushed to quay (default: false)
```
bparees, line 21: seems like this is also rebuilding + pushing the operator image?

in fact it seems like that is the primary distinction this argument introduces?  so i'm wondering if the argument name and description are correct.

hack/publish_package_builds.sh
```diff
@@ -138,12 +167,8 @@ fi
...
 echo "Building catalog image with  ${TARGET_BUNDLE_IMAGE}"
 
-cat << EOF > "${CATALOG_FILE}"
----
-defaultChannel: preview
-name: lightspeed-operator
-schema: olm.package
-EOF
+#Append ./hack/operator.yaml to lightspeed-catalog/index.yaml
+cat "${CATALOG_INTIAL_FILE}" > "${CATALOG_FILE}"
```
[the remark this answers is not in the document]
xrajesh: fixed.

hack/publish_package_builds.sh
xrajesh: We dont need it  here . Removed .

hack/publish_package_builds.sh
xrajesh:
```
--rebuild-all:
    If set, --operator-image and --publish options are required. The bundle and catalog images are built and pushed to quay (default: false)
```
By rebuild-all - I want a option to pull the operator github reposirtory and make a change and
1. Build the operator using the PullSpec --operator-image 
2. Build the bundle , by completely regenerating the bundle artifacts
3. Build the catalog image using the bundle image.
 If the name is confusing I am open for suggestions @bparees . For example , I built catalog with specific UI image and latest ols api - with my current operator changes like this.
```
./hack/publish_package_builds.sh -o xrajesh  --rebuild-all --operator-image quay.io/xrajesh/lightspeed-operator:latest -t 0.0.3 -p --console-image quay.io/openshift/lightspeed-console-plugin@sha256:a320bc76b7772d077b75d76c5768ed63dae0de1430dba228d72ddd42df150673 --ols-image quay.io/openshift/lightspeed-service-api:latest
```

bparees: if nothing else the description is wrong since w/ this arg the operator image is also rebuilt (and pushed?), since you run this:


```
 make docker-build docker-push VERSION="${VERSION}" IMG="${OPERATOR_IMAGE}"
 ```

but the description says nothing about that.


the description also claims that "--publish" needs to be set when using --rebuild-all, but it's not clear why and the script doesn't check/enforce that.

I think I would instead introduce an arg for "--build-operator" that causes the operator image to be rebuilt.

and then tie whether or not the operator image gets pushed to the --publish flag(only pushing it if it was also built), or introduce a separate --publish-operator flag.

bparees:
/lgtm
/approve

the --rebuild-all flag can be reworked post merge, this will unblock https://github.com/openshift/lightspeed-operator/pull/86 so we can get metric reporting to telemetry.

PR #115, opened 2024-04-24
Enable collection

onmete:
## Description

Add a user data collection config entry.

## Type of change

- [x] New feature


bparees:
/lgtm

@xrajesh ptal

PR #118, opened 2024-04-25
 add PolicyRule to get for pull-secret

xrajesh:
## Description

 Added PolicyRule to get for pull-secret present in namespace openshift-config

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/internal/controller/ols_app_server_assets.go
+++ b/internal/controller/ols_app_server_assets.go
@@ -61,6 +61,11 @@ func (r *OLSConfigReconciler) generateSARClusterRole(cr *olsv1alpha1.OLSConfig) 
 				Resources: []string{"clusterversions"},
 				Verbs:     []string{"get"},
 			},
+			{
+				APIGroups: []string{""},
+				Resources: []string{"secrets"},
+				Verbs:     []string{"get"},
+			},
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


internal/controller/ols_app_server_assets.go
```diff
@@ -61,6 +61,11 @@ func (r *OLSConfigReconciler) generateSARClusterRole(cr *olsv1alpha1.OLSConfig)
 				Resources: []string{"clusterversions"},
 				Verbs:     []string{"get"},
 			},
+			{
+				APIGroups: []string{""},
+				Resources: []string{"secrets"},
+				Verbs:     []string{"get"},
```
bparees, line 67: constrain this by setting the `resourceNames` value to just include the pull-secret resource name.

but ultimately i think the correct way to do this would be to introduce a namespace specific rolebinding that only grants this permission to OLS in the openshift-config NS (as well as role only granting it against the `pull-secret` resource name).

that can be a followup since it will be more work to implement logic for the new rolebinding creation, as well as creating a separate role that can be bound just at the NS level instead of cluster-bound.

PR #121, opened 2024-04-29
OLS-346: periodic reconciliation

xrajesh:
## Description

This PR allows the operator to reconcile even when no event is received. This prevents missing reonciliation and leaves the OLS stack in an inconsistent state.

Rebased https://github.com/openshift/lightspeed-operator/pull/90 . 

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

Related Issue # [OLS-346](https://issues.redhat.com//browse/OLS-346)
- Closes #

## Testing

### DIFF
--- a/internal/controller/constants.go
+++ b/internal/controller/constants.go
@@ -5,7 +5,7 @@ const (
 	// OLSConfigName is the name of the OLSConfig Custom Resource
 	OLSConfigName = "cluster"
 	// DefaultReconcileInterval is the default interval for reconciliation
-	DefaultReconcileInterval = 300
+	DefaultReconcileInterval = 120
 
 	/*** application server configuration file ***/
 	// OLSConfigName is the name of the OLSConfig configmap
--- a/internal/controller/constants.go
+++ b/internal/controller/constants.go
@@ -4,6 +4,8 @@ const (
 	/*** Operator Settings ***/
 	// OLSConfigName is the name of the OLSConfig Custom Resource
 	OLSConfigName = "cluster"
+	// DefaultReconcileInterval is the default interval for reconciliation
+	DefaultReconcileInterval = 300
 
 	/*** application server configuration file ***/
 	// OLSConfigName is the name of the OLSConfig configmap
--- a/internal/controller/olsconfig_controller.go
+++ b/internal/controller/olsconfig_controller.go
@@ -168,6 +168,7 @@ func (r *OLSConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, nil
 	}
 	r.NextReconcileTime = time.Now().Add(r.Options.ReconcileInterval)
+	r.logger.Info("Next automatic reconciliation scheduled at", "nextReconcileTime", r.NextReconcileTime)
 	return ctrl.Result{RequeueAfter: r.Options.ReconcileInterval}, nil
 }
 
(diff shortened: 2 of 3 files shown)


internal/controller/constants.go
```diff
@@ -4,6 +4,8 @@ const (
 	/*** Operator Settings ***/
 	// OLSConfigName is the name of the OLSConfig Custom Resource
 	OLSConfigName = "cluster"
+	// DefaultReconcileInterval is the default interval for reconciliation
+	DefaultReconcileInterval = 300
```
bparees, line 8: i think something more like 2 hours is typical for relisting/forcing reconciliation.  Did you base this value on another operator?

internal/controller/olsconfig_controller.go
```diff
@@ -159,7 +163,12 @@ func (r *OLSConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	// Update status condition for Custom Resource
 	r.updateStatusCondition(ctx, olsconfig, typeCRReconciled, true, "Custom resource successfully reconciled", nil)
 
-	return ctrl.Result{}, nil
+	// Requeue if no reconciliation is scheduled in future.
+	if r.NextReconcileTime.After(time.Now()) {
+		return ctrl.Result{}, nil
+	}
+	r.NextReconcileTime = time.Now().Add(r.Options.ReconcileInterval)
+	return ctrl.Result{RequeueAfter: r.Options.ReconcileInterval}, nil
```
bparees, line 171: so we're requeuing the event despite having processed it successfully, just to trigger a future reconciliation?

bparees (commented):
this seems to cover re-listing/reconciling the olsconfig CR periodically.

What about all the other resources the operator watches/all the resources that affect what the operator generates?  (secrets, deployments, etc).

i'm a little surprised you have to implement this yourself, also, traditionally i thought you could just configure a relist interval on an Informer, at least for resources you are watching.

internal/controller/constants.go
xrajesh: We only had this as starting point. I am good with 2 hours - I will udpate

internal/controller/olsconfig_controller.go
xrajesh: Correct. Just keep the periodic reconcile going - only for a case if  we missed a event.

xrajesh:
this seems to cover re-listing/reconciling the olsconfig CR periodically. 
-- Yes . We will ensure that we are running with desired state specified in CR - even if we miss a event , every 300 seconds.

For other resources, we rely on reconciling the owner object . https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/builder#Builder.Owns

bparees:
> For other resources, we rely on reconciling the owner object . https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/builder#Builder.Owns

does that come with periodic relisting behavior?

xrajesh:
>does that come with periodic relisting behavior?
I could not find a way to periodic reconcile for the Owned resources. But For those resources - CREATE is not a concern because API Server will not let the user create a resource with same name. DELETE is not a concern because the periodic OLSCONFIG will create during reconcile. Only UPDATE is a concern ( if we missed the event) . I feel its a rare case. WDYT @bparees

PR #150, opened 2024-05-30
OLS-579:  use MTLS on metric endpoint, replacing kube-rbac-proxy

raptorsun:
## Description

use Mutual TLS on metric endpoint of the operator, replacing the kube-rbac-proxy and the RBAC settings to scrape metrics from the operator.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing


raptorsun:
/hold
adding tests

raptorsun:
https://github.com/openshift/release/pull/52728 should fix the bundle-e2e test problem.
/unhold

config/rbac/role.yaml
```diff
@@ -174,3 +174,18 @@ rules:
   - roles
   verbs:
   - '*'
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: manager-role
+  namespace: openshift-monitoring
```
bparees, line 182: @raptorsun so are you saying that even though this role has the NS defined as "openshift-monitoring", when we actually install the operator, the role is being created in the "openshift-lightspeed" NS instead?

and that you think the solution(workaround) is to make this a clusterrole instead?

bparees (commented):
does the operator test suite include a test that confirms that the operator metrics are successfully scraped by prometheus?  (in other words, a test that queries prometheus itself for the metrics you expect the operator to be reporting)?

that is the most critical test to have for a change like this, to ensure that prometheus is still succesfully scraping metrics from the operator after the change.

raptorsun:
> does the operator test suite include a test that confirms that the operator metrics are successfully scraped by prometheus? (in other words, a test that queries prometheus itself for the metrics you expect the operator to be reporting)?
> 
> that is the most critical test to have for a change like this, to ensure that prometheus is still succesfully scraping metrics from the operator after the change.

I pushed a new version adding a test to query Prometheus for metrics scrapped from OLS operator. We can verify whether the metrics are saved into Prometheus. Thanks for reminding me adding the test :)

bparees:
@raptorsun on this one i think we are waiting on a summary/explanation of all the cert keys and cert CAs involved so we can understand:

1) where does the cert key that prometheus uses come from?
2) what CA signs that cert?
3) how does prometheus know to supply that cert key when scraping the operator metrics endpoint?
4) how does the operator know to trust that cert key (presumably by trusting the CA in (2))?
5) do any of those certs+CAs have any relationship to the cert key+CA that is used by the metrics endpoint for providing TLS?  Looks like it's probably a different/unrelated key/CA pair (created by the service signing CA system)

raptorsun:
1. Where does the cert key that prometheus uses come from?

Prometheus uses a certificate stored in the secret "metrics-client-certs" in the namespace “openshift-monitoring”. 

2.what CA signs that cert?

This client certificate is requested by the Cluster Monitoring Operator through a CertificateSigningRequest. The CA that signs this certificate is the CA held by the kube apiserver. The CA certificate is saved in the configmap “extension-apiserver-authentication” in the namespace “kube-system”.
Cluster Monitoring Operator keeps a copy of the CA in the configmap “metrics-client-ca” in the namespace “openshift-monitoring”

3. How does prometheus know to supply that cert key when scraping the operator metrics endpoint?
In the ServiceMonitor we specify the TLS settings such as CA, client certificate, private key.
```yaml
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: controller-manager-metrics-monitor
spec:
  endpoints:
    - path: /metrics
      port: metrics
      scheme: https
      tlsConfig:
        insecureSkipVerify: false
        caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
        certFile: /etc/prometheus/secrets/metrics-client-certs/tls.crt
        keyFile: /etc/prometheus/secrets/metrics-client-certs/tls.key
        serverName: lightspeed-operator-controller-manager-service.openshift-lightspeed.svc
  selector:
    matchLabels:
      control-plane: controller-manager
```

4. How does the operator know to trust that cert key (presumably by trusting the CA in (2))?
The Lightspeed Operator extract the certificate of the CA signing prometheus’ client certificate from the configmap  “metrics-client-ca” in the namespace “openshift-monitoring”. Then it uses the CA certificate to check client’s certificate. 

5. do any of those certs+CAs have any relationship to the cert key+CA that is used by the metrics endpoint for providing TLS? Looks like it's probably a different/unrelated key/CA pair (created by the service signing CA system)

On the operator’s metrics endpoint, the CA is the one used to sign certificates for services, whose certificate is saved in a configmap with the annotation “service.beta.openshift.io/inject-cabundle: "true"”. The operator’s TLS certificate and private key is requested by adding an annotation “service.beta.openshift.io/serving-cert-secret-name: controller-manager-tls” in its Service. The CA signing the operator’s  client certificate is the CA whose certificate is injected into the configmap just mentioned.

Both CAs are different, one is for authentication of service endpoints and the other is for authentication of clients, in our case, a service account.  They are not related at all

. 
MTLS

bundle/manifests/lightspeed-operator.clusterserviceversion.yaml
```diff
@@ -324,9 +312,11 @@ spec:
             spec:
               containers:
               - args:
-                - --health-probe-bind-address=:8081
-                - --metrics-bind-address=127.0.0.1:8080
                 - --leader-elect
+                - --health-probe-bind-address=:8081
+                - --metrics-bind-address=:8080
```
bparees, line 317: if the metrics endpoint is now directly TLS encrypted, we should move it to port 8443 to reflect that, 8080 should be used for unencrypted endpoints.

config/manager/manager.yaml
```diff
@@ -67,6 +67,10 @@ spec:
         - /manager
         args:
         - --leader-elect
+        - "--health-probe-bind-address=:8081"
+        - "--metrics-bind-address=:8080"
```
bparees, line 71: 8443?

config/manager/manager.yaml
```diff
@@ -87,6 +91,14 @@ spec:
             port: 8081
           initialDelaySeconds: 5
           periodSeconds: 10
+        ports:
+          - containerPort: 8080
```
bparees, line 95: 8443?

config/rbac/role.yaml
bparees: need an answer/resolution on this thread

test/e2e/metrics_test.go
```diff
@@ -0,0 +1,113 @@
...
+		client, err = GetClient(nil)
+		Expect(err).NotTo(HaveOccurred())
+		By("Deleting the OLSConfig CR")
+		Expect(cr).NotTo(BeNil())
+		err = client.Delete(cr)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("should have operator metrics in Prometheus", func() {
+		By("verify Prometheus is scrapping its own metrics")
```
bparees, line 92:
```suggestion
		By("verify Prometheus is scraping the ols operator metrics")
```

test/e2e/metrics_test.go
```diff
@@ -0,0 +1,113 @@
...
+		err = prometheusClient.WaitForQueryReturnGreaterEqualOne("prometheus_build_info")
+		Expect(err).NotTo(HaveOccurred())
+
+		By("wait for prometheus scrape metrics from operator")
+		time.Sleep(60 * time.Second)
```
bparees, line 97: better to retry until success (for up to a minute or whatever timeout you want).  Otherwise you're forcing this test to always take at least a minute, even if the metric data might be available sooner.

raptorsun:
e2e bundle test is fixed https://github.com/openshift/release/pull/53287

PR #153, opened 2024-06-05
OLS-500: disable data collection when telemetry is off

raptorsun:
## Description
disable data collection when telemetry is off. 

Telemetry enablement is determined by the presence of the telemetry pull secret: the presence of the field '.auths."cloud.openshift.com"' indicates that telemetry is enabled.
Use this command to check in an Openshift cluster
```oc get secret/pull-secret -n openshift-config --template='{{index .data ".dockerconfigjson" | base64decode}}' | jq '.auths."cloud.openshift.com"'```

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue # 
- Closes # OLS-500

## Testing

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -44,6 +44,16 @@ rules:
   - list
   - watch
 - apiGroups:
+  - ""
+  resourceNames:
+  - pull-secret
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
   - ols.openshift.io
   resources:
   - olsconfigs
--- a/internal/controller/constants.go
+++ b/internal/controller/constants.go
@@ -111,4 +111,9 @@ const (
 
 	/*** watchers ***/
 	WatcherAnnotationKey = "ols.openshift.io/watcher"
+	// TelemetryPullSecretNamespace "openshift-config" contains the telemetry pull secret to determine the enablement of telemetry
+	// #nosec G101
+	TelemetryPullSecretNamespace = "openshift-config"
+	// TelemetryPullSecretName is the name of the secret containing the telemetry pull secret
+	TelemetryPullSecretName = "pull-secret"
 )
--- a/internal/controller/ols_app_server_assets_test.go
+++ b/internal/controller/ols_app_server_assets_test.go
@@ -29,6 +29,7 @@ var _ = Describe("App server assets", func() {
 	var rOptions *OLSConfigReconcilerOptions
 	var secret *corev1.Secret
 	defaultVolumeMode := int32(420)
+
 	Context("complete custom resource", func() {
 		BeforeEach(func() {
 			rOptions = &OLSConfigReconcilerOptions{
@@ -201,10 +202,14 @@ var _ = Describe("App server assets", func() {
 		})
 
 		It("should generate the OLS deployment", func() {
+			By("generate full deployment when telemetry pull secret exists")
+			createTelemetryPullSecret()
+
 			dep, err := r.generateOLSDeployment(cr)
 			Expect(err).NotTo(HaveOccurred())
 			Expect(dep.Name).To(Equal(OLSAppServerDeploymentName))
 			Expect(dep.Namespace).To(Equal(OLSNamespaceDefault))
+			// application container
 			Expect(dep.Spec.Template.Spec.Containers[0].Image).To(Equal(rOptions.LightspeedServiceImage))
 			Expect(dep.Spec.Template.Spec.Containers[0].Name).To(Equal("lightspeed-service-api"))
 			Expect(dep.Spec.Template.Spec.Containers[0].Resources).ToNot(BeNil())
@@ -243,6 +248,7 @@ var _ = Describe("App server assets", func() {
 					MountPath: "/app-root/ols-user-data",
 				},
 			}))
+			// telemetry container
 			Expect(dep.Spec.Template.Spec.Containers[1].Image).To(Equal(rOptions.LightspeedServiceImage))
 			Expect(dep.Spec.Template.Spec.Containers[1].Name).To(Equal("lightspeed-service-user-data-collector"))
 			Expect(dep.Spec.Template.Spec.Containers[1].Resources).ToNot(BeNil())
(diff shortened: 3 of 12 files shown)

### CODE REVIEW

File: bundle/manifests/lightspeed-operator-lightspeed-operator-controller-manager_v1_serviceaccount.yaml [hunk 1]
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: ServiceAccount
```
bparees, line 2: why is this PR contributing a new service account definition?

internal/controller/constants.go
```diff
@@ -98,4 +98,9 @@ const (
 
 	/*** watchers ***/
 	WatcherAnnotationKey = "ols.openshift.io/watcher"
+	// TelemetryPullSecretNamespace "openshift-config" contains the telemetry pull secret to determine the enablement of telemetry
+	// #nosec G101
+	TelemetryPullSecretNamespace = "openshift-config"
+	// TelemetryPullSecretName is the name of the secret containing the telemetry pull secret
+	TelemetryPullSecretName = "pull-secret"
```
bparees, line 105: don't you need a new role+rolebinding to grant the operator permission to read/watch this resource?

bparees (commented):
i'm a little confused about what seem to be unrelated changes to the service account definitions (maybe caused by the controller runtime bump?)

bundle/manifests/lightspeed-operator-lightspeed-operator-controller-manager_v1_serviceaccount.yaml
raptorsun: ohh, I should have deleted it. it was a residue from a test. Thanks!

internal/controller/constants.go
raptorsun: the role is here https://github.com/openshift/lightspeed-operator/pull/153/files#diff-6d4c4e29ec096149781536bb9118fde8242c1901df764e3eda3019ac8503d6c7R82-R93

raptorsun: during the installation from bundle, a clusterrole is created with more than the required access rights on secret.
It seems more rights  granted than we wanted though. 
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  creationTimestamp: "2024-06-11T08:35:46Z"
  labels:
    olm.managed: "true"
    olm.owner: lightspeed-operator.v0.0.1
    olm.owner.kind: ClusterServiceVersion
    olm.owner.namespace: openshift-lightspeed
    olm.permissions.hash: 4ocN4MAqabrmojboiME86jksllfmx47XTsRpBP
    operators.coreos.com/lightspeed-operator.openshift-lightspeed: ""
  name: lightspeed-operator.v0.0-5Aqfwi9a4ZKSKdIsDy4AsLMj5iQ6GUFBB15dIZ
  resourceVersion: "42573"
  uid: 021a457a-e621-4fbb-8527-d0a5f94059b6
rules:
- apiGroups:
  - ""
  resources:
  - configmaps
  verbs:
  - get
  - list
  - watch
  - create
  - update
  - patch
  - delete
- apiGroups:
  - coordination.k8s.io
  resources:
  - leases
  verbs:
  - get
  - list
  - watch
  - create
  - update
  - patch
  - delete
- apiGroups:
  - ""
  resources:
  - events
  verbs:
  - create
  - patch
- apiGroups:
  - apps
  resources:
  - deployments
  verbs:
  - create
  - delete
  - get
  - list
  - patch
  - update
  - watch
- apiGroups:
  - ""
  resources:
  - configmaps
  verbs:
  - create
  - delete
  - get
  - list
  - patch
  - update
  - watch
- apiGroups:
  - ""
  resources:
  - secrets
  verbs:
  - create
  - delete
  - get
  - list
  - patch
  - update
  - watch
- apiGroups:
  - ""
  resources:
  - serviceaccounts
  verbs:
  - create
  - delete
  - get
  - list
  - patch
  - update
  - watch
- apiGroups:
  - ""
  resources:
  - services
  verbs:
  - create
  - delete
  - get
  - list
  - patch
  - update
  - watch
- apiGroups:
  - monitoring.coreos.com
  resources:
  - prometheusrules
  verbs:
  - create
  - delete
  - get
  - list
  - patch
  - update
  - watch
- apiGroups:
  - monitoring.coreos.com
  resources:
  - servicemonitors
  verbs:
  - create
  - delete
  - get
  - list
  - patch
  - update
  - watch
- apiGroups:
  - rbac.authorization.k8s.io
  resources:
  - rolebindings
  - roles
  verbs:
  - '*'
- apiGroups:
  - ""
  resources:
  - namespaces
  verbs:
  - get
  - list
  - watch
```

raptorsun:
fix for the version conflict error in e2e test https://github.com/openshift/lightspeed-operator/pull/159

config/rbac/role.yaml
```diff
@@ -79,6 +79,21 @@ rules:
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
+metadata:
+  name: manager-role
```
bparees, line 83: there are several roles named "manager-role" in this file.  Please give these roles names that make it clearer what purpose they serve(e.g. cluster-pull-secret-reader").  Otherwise admins are just going to see a "manager-role" role in the "openshift-config" NS and have no idea why it exists/what uses it.

internal/controller/constants.go
bparees: > the role is here https://github.com/openshift/lightspeed-operator/pull/153/files#diff-6d4c4e29ec096149781536bb9118fde8242c1901df764e3eda3019ac8503d6c7R82-R93

what about the rolebinding that uses the role?

config/rbac/role.yaml
raptorsun: these roles are generated by kubebuilder that translate the RBAC settings such as `// +kubebuilder:rbac:groups=core,namespace=openshift-config,resources=secrets,verbs=get;list;watch` to the roles.

bparees: looks like it's controlled by the makefile, so we can change it:

https://github.com/kubernetes-sigs/kubebuilder/discussions/2297#discussioncomment-3766946

https://github.com/openshift/lightspeed-operator/blob/e76f8bb870bb1564aa33f55873d70e3fbca516b2/Makefile#L102

though i don't see a way to make it unique per role unfortunately.

bparees:
> Now the operator still relies on clusterrole to access resource in other namespaces.

i think that's ok, but probably worth a question to the OLM team about whether there is something we're missing that would allow us to define a role in another NS and then bind to it.

config/rbac/role.yaml
```diff
@@ -79,6 +79,21 @@ rules:
...
+  namespace: openshift-config
```
bparees, line 84: so per your comment here:

https://github.com/openshift/lightspeed-operator/pull/153#issuecomment-2162690983

does this role need to be turned into a clusterrole?

internal/controller/suite_test.go
```diff
@@ -108,6 +108,15 @@ var _ = BeforeSuite(func() {
 	err = k8sClient.Create(ctx, ns)
 	Expect(err).NotTo(HaveOccurred())
 
+	By("Create the namespace openshift-config")
+	ns = &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "openshift-config",
+		},
+	}
+	err = k8sClient.Create(ctx, ns)
```
bparees, line 117: this NS should exist on every OCP cluster, you should not have to create it.....did you have some issue?

bparees: ah, ok.  sorry, was lacking context

config/rbac/role.yaml
raptorsun: that's generated by controller-gen. 
If we remove the namespace limit [here](https://github.com/openshift/lightspeed-operator/pull/153/files#diff-67e7ce7f3b59749b56b35a828cd9e81ce48fcda60db9821752fe3f8b04780051R81-R82), these permissions will be added to the clusterrole of the operator service account. 

But for some reason yet to find out with the OLM team, the operator service account has already access to configmaps in other namespaces due to cluster role created during the bundle installation.

bparees:
@raptorsun i think the logic in this PR is ok, but we need to get to the bottom of the questions about the roles that are being created and why.

It sounds like ideally we want to create the role in the openshift-config NS, but we don't know if that's possible?  We could live with it being a clusterrole if necessary.

And also a clusterrole is being created that we aren't even asking for and which is more expansive than we really want?

raptorsun:
> @raptorsun i think the logic in this PR is ok, but we need to get to the bottom of the questions about the roles that are being created and why.
> 
> It sounds like ideally we want to create the role in the openshift-config NS, but we don't know if that's possible? We could live with it being a clusterrole if necessary.
> 
> And also a clusterrole is being created that we aren't even asking for and which is more expansive than we really want?

After consulting a operator-sdk person, I find the `operator-sdk` and OLM does not create role and rolebindings in other namespaces than the namespace where the operator is deployed. So I put the required resource access into the cluster role instead of the namespaced role.

config/rbac/role.yaml
```diff
@@ -35,6 +35,14 @@ rules:
   - delete
   - get
   - update
+- apiGroups:
+  - ""
+  resources:
+  - secrets
```
bparees, line 41: Since we are binding this clusterrole as a clusterrolebinding, that means the OLS service account gets all these permissions in all namespaces.

can we constrain this permission to just the `resourceName` we care about?  Or will that break the list/watch permission since list/watch isn't specific to a particular resource name?

raptorsun: Yes, we can constrain the resource name to `pull-secret`. A new version is pushed with this modification.

internal/controller/ols_app_server_assets_test.go
```diff
@@ -300,6 +306,89 @@ var _ = Describe("App server assets", func() {
 				},
 			}))
 			Expect(dep.Spec.Selector.MatchLabels).To(Equal(generateAppServerSelectorLabels()))
+
+			By("generate deployment without data collector when telemetry pull secret does not exist")
+			deleteTelemetryPullSecret()
```
bparees, line 311: the expected scenario will not be that the entire pullsecret is deleted, it will be that the pull secret doesn't contain an auth entry for `cloud.openshift.com`, so is that scenario being explicitly tested somewhere?

raptorsun: I added a test that the pull-secret exists without the entry for `cloud.openshift.com` in the new commit. Please have a look :)

bparees:
needs a rebase now, but
/approve

raptorsun:
rebased now :)  @bparees

PR #179, opened 2024-06-25
Update README.md to include information about other LLMs

AlexonOliveiraRH:
Adding an explanation about the compatible LLMs.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ A Kubernetes operator for managing [Red Hat OpenShift Lightspeed](https://github
 You'll need an OpenShift cluster to run against.
 
 > [!IMPORTANT]
-> Officially, the Operator only supports OpenAI and BAM as large language model (LLM) providers, since we do not test for other models, but technically, if you have an OpenAI API compatible model server (Ollama, VLLM, MLX), it should work.
+> Officially, the Operator only supports OpenAI, Azure OpenAI and WatsonX as large language model (LLM) providers, but technically, if you have an OpenAI API compatible model server (Ollama, VLLM, MLX), it should work.
 
 ### Running on the cluster
 
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ A Kubernetes operator for managing [Red Hat OpenShift Lightspeed](https://github
 You'll need an OpenShift cluster to run against.
 
 > [!IMPORTANT]
-> The Operator only supports OpenAI and BAM as large language model (LLM) providers.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,7 +7,7 @@ A Kubernetes operator for managing [Red Hat OpenShift Lightspeed](https://github
 You'll need an OpenShift cluster to run against.
 
 > [!IMPORTANT]
-> The Operator only supports OpenAI and BAM as large language model (LLM) providers.
+> Officially, the Operator only supports OpenAI and BAM as large language model (LLM) providers, since we do not test for other models, but technically, if you have an OpenAI API compatible model server (Ollama, VLLM, MLX), it should work.
```
bparees, line 10: BAM shouldn't really be mentioned since that's an IBM-internal research thing no one else has access to.

Also we supposer `Azure OpenAI` and `WatsonX` in addition to `OpenAI` as a provider.

bparees: > since we do not test for other models

it's a nuance, but model!=provider.

We support(and test with) a certain set of providers, and for those providers, we support+test with a certain set of models.  (For example you can get an instance of llama via WatsonX, but we do not test against llama.  In theory you could configure OLS to use it, we have no idea what might happen)

AlexonOliveiraRH: @bparees BAM was originally mentioned. If that's the case, we should omit this.
We can add info about the other mentioned providers, no problem.
About the `since we do not test for other models`, we can remove this sentence too.

AlexonOliveiraRH: @bparees done, adjusted.

bparees: > @bparees BAM was originally mentioned. If that's the case, we should omit this.

yup, i understand it wasn't your contribution but if we're going to update the text i'd like to see all the existing issues w/ it cleaned up

AlexonOliveiraRH: > > @bparees BAM was originally mentioned. If that's the case, we should omit this.
> 
> yup, i understand it wasn't your contribution but if we're going to update the text i'd like to see all the existing issues w/ it cleaned up

Totally agree. I've made the required adjustment. Can you check it, please?

bparees: looks ok to me, will let @xrajesh sign off on it

PR #227, opened 2024-07-10
OLS-650: Add extra certs if defined

onmete:
## Description

Add extra certs if defined - WIP

## Type of change

- [x] New feature


internal/controller/ols_app_server_deployment.go
```diff
@@ -129,6 +136,31 @@ func (r *OLSConfigReconciler) generateOLSDeployment(cr *olsv1alpha1.OLSConfig) (
...
+		for key:= range extraCertsCM.Data {
+			volumeName := OLSExtraCertsVolumeName + "-" + key
+			volume := corev1.Volume{
+				Name: volumeName,
+				VolumeSource: corev1.VolumeSource{
+					ConfigMap: &corev1.ConfigMapVolumeSource{
+						LocalObjectReference: corev1.LocalObjectReference{
+							Name: configMapName,
+						},
+						Items: []corev1.KeyToPath{
```
onmete, line 151: This is probably wrong, I think I need to use the actual data that are stored in CM under the key.

api/v1alpha1/olsconfig_types.go
```diff
@@ -102,6 +102,9 @@ type OLSSpec struct {
 	// User data collection switches
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="User Data Collection"
 	UserDataCollection UserDataCollectionSpec `json:"userDataCollection,omitempty"`
+	// extra certs - list of paths
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Extra Certs"
+	ExtraCerts []string `json:"extraCerts,omitempty"`
```
raptorsun, line 107: We should put either name of configmaps/secrets here, or the content of the certificates. 
The file path as well as the app server config file should be generated by the operator at runtime.

1. If we put into this field the certificate itsself, operator creates the configmap holding these certificates, mount it to the app server deployment and add mouted path to the config file. 
2. If we put a configmap/secret name, we have these options:
    A. assume the certificate key name in the configmap/secret 
    B. we assume all the keys in the configmap/secret is certificates
    C. Add another field for the key in the configmap/secret to locate the certificate.

internal/controller/ols_app_server_assets.go
```diff
@@ -205,6 +205,7 @@ func (r *OLSConfigReconciler) generateOLSConfigMap(ctx context.Context, cr *olsv
 			TranscriptsDisabled: cr.Spec.OLSConfig.UserDataCollection.TranscriptsDisabled,
 			TranscriptsStorage:  "/app-root/ols-user-data/transcripts",
 		},
+		ExtraCerts: cr.Spec.OLSConfig.ExtraCerts,
```
raptorsun, line 208: operator shall generate this section of config instead of a bare copy from CR.

internal/controller/ols_app_server_deployment.go
raptorsun: If we set all keys of configmap available in this volume, we can omit the `Items`.
Reference https://kubernetes.io/docs/concepts/storage/volumes/#configmap

internal/controller/ols_app_server_deployment.go
```diff
@@ -386,3 +438,18 @@ func (r *OLSConfigReconciler) dataCollectorEnabled(cr *olsv1alpha1.OLSConfig) (b
 	}
 	return configEnabled && telemetryEnabled, nil
 }
+
+// extraCertsDefined checks if the ConfigMap named ols-extra-ca exists and returns it if so
+func (r *OLSConfigReconciler) extraCertsDefined(namespace string) (bool, *corev1.ConfigMap, error) {
+	configMap := &corev1.ConfigMap{}
+	err := r.Client.Get(context.TODO(), client.ObjectKey{
+		Name:      "ols-extra-ca",
```
raptorsun, line 446: I suggest we leave the configmap name configurable in the OLSConfig CR. It is difficult for users to memorize these constants.

raptorsun:
we may need declare additional writable emphemeral volume for the app server container to host the combined certificates as mentioned in this PR. 
https://github.com/openshift/lightspeed-service/pull/1216/files#diff-34cde5e49b54d9b67952c80d3580b6c421ebcf16097996e215aef54e44f78562R73

onmete:
Closing.

PR #323, opened 2024-08-08
[OLs v0.1.1 ] added upgrade graph with skipRange

xrajesh:
## Description

Added skipRange for upgrade 
Picked the service SHA which we released.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing


lightspeed-catalog-4.16/index.yaml
```diff
@@ -62,3 +62,4 @@ package: lightspeed-operator
 name: preview
 entries:
   - name: lightspeed-operator.v0.1.1
+    skipRange: ">=0.1.0 <0.1.1"
```
joshuawilson, line 65: Does this mean, skip "0.1.0"?
Why not just use "=0.1.0"?

xrajesh: From the doc 
```
The entry specifying the skipRange will be presented as a direct (one hop) update to any version from that package within that range. The versions in this range do not need to be in the catalog in order for bundle addition to be successful. We recommend avoiding using unbound ranges such as <1.0.3.
```

PR #379, opened 2024-09-05
OLS-886 snapshot_to_catalog.sh produce 2 channels with multiple bundle versions

raptorsun:
## Description

The script snapshot_to_catalog.sh produce 2 channels `alpha` and `preview` into the catalog.
The channel `preview` has the latest version and skips all older versions.
The channel `alpha` has all versions provided to the script, one replacing the closest older one.
Here is a catalog generated from command `./hack/snapshot_to_catalog.sh -s ols-cq8sl,ols-mdc8x`
More detail see [this PR](https://github.com/openshift/lightspeed-operator/pull/379)
```yaml
---
defaultChannel: preview
icon:
  base64data: ...
name: lightspeed-operator
schema: olm.package
---
image: registry.redhat.io/openshift-lightspeed-beta/lightspeed-operator-bundle@sha256:fd3849cb65fee7586b261c8c67336d1d0e5d98e89ae10f87f7ca7945d585b0ff
name: lightspeed-operator.v0.1.3
package: lightspeed-operator
properties:
  - type: olm.gvk
    value:
      group: ols.openshift.io
      kind: OLSConfig
      version: v1alpha1
  - type: olm.package
    value:
      packageName: lightspeed-operator
      version: 0.1.3
  - type: olm.bundle.object
    value:
      data:  ...
relatedImages:
  - name: lightspeed-service-api
    image: registry.redhat.io/openshift-lightspeed-beta/lightspeed-service-api-rhel9@sha256:9f4b71d0bfd0cdda5f872adf38a792eb7a0f0cba9097496cb1843b05f6f3e890
  - name: lightspeed-console-plugin
    image: registry.redhat.io/openshift-lightspeed-beta/lightspeed-console-plugin-rhel9@sha256:ebe5b5bbe3af1d6089667a001686ad152084a684f021abc52ec2884b7bae790b
  - name: lightspeed-operator
    image: registry.redhat.io/openshift-lightspeed-beta/lightspeed-rhel9-operator@sha256:af95acfcb1e685a8023cad68707e60cb81de2e117ce0d438e6481dbee3b60841
  - name: lightspeed-operator-bundle
    image: registry.redhat.io/openshift-lightspeed-beta/lightspeed-operator-bundle@sha256:fd3849cb65fee7586b261c8c67336d1d0e5d98e89ae10f87f7ca7945d585b0ff
schema: olm.bundle
---
[...]

PR #381, opened 2024-09-05
Example Output for PR379

raptorsun:
Example output of [PR#379](https://github.com/openshift/lightspeed-operator/pull/379)


lightspeed-catalog-4.15/index.yaml
```diff
@@ -63,3 +114,12 @@ name: preview
...
     skipRange: ">=0.1.0 <0.1.4"
+---
+schema: olm.channel
+package: lightspeed-operator
+name: alpha
+entries:
+  - name: lightspeed-operator.v0.1.3
+    skipRange: ">=0.1.0 <0.1.3"
+  - name: lightspeed-operator.v0.1.4
+    replaces: lightspeed-operator.v0.1.3
```
xrajesh, line 125: What are we gaining in this definition  than just specifying `skipRange: ">=0.1.0 <0.1.4" @raptorsun

raptorsun: if we pass only on snapshot, it will list a single item and skip all older versions

xrajesh:
@raptorsun This is just a example output right ? we dont have to merge this .

raptorsun:
> @raptorsun This is just a example output right ? we dont have to merge this .

yes, this is example, we do not neet to merge this PR.

raptorsun:
/hold
just an example

raptorsun:
/close
no longer needed

PR #728, opened 2025-05-05
OLS-1686: OLS should persist postgres data

syedriko:
## Description

Added a PVC for the Postgres data volume

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes # [OLS-1686](https://issues.redhat.com//browse/OLS-1686)

## Testing


raptorsun:
shall we disable the persistent volume claim during the E2E test? test clusters do not seem to have persistent volumes.
I think we should have an option in the OLSConfig CR for the user to opt out the persistency.

xrajesh:
> shall we disable the persistent volume claim during the E2E test? test clusters do not seem to have persistent volumes. I think we should have an option in the OLSConfig CR for the user to opt out the persistency.

Storage setting is optional right - user need not do anything to opt out ? @syedriko ?

syedriko:
> > shall we disable the persistent volume claim during the E2E test? test clusters do not seem to have persistent volumes. I think we should have an option in the OLSConfig CR for the user to opt out the persistency.
> 
> Storage setting is optional right - user need not do anything to opt out ? @syedriko ?

As it's written now, persistence is mandatory, as I'd argue it should be. There're two knobs - one for PVC size and another one for storage class and they are optional with detaults, but persistence is always there. I was actually hoping to add automated e2e tests for persistence...

raptorsun:
> > > shall we disable the persistent volume claim during the E2E test? test clusters do not seem to have persistent volumes. I think we should have an option in the OLSConfig CR for the user to opt out the persistency.
> > 
> > 
> > Storage setting is optional right - user need not do anything to opt out ? @syedriko ?
> 
> As it's written now, persistence is mandatory, as I'd argue it should be. There're two knobs - one for PVC size and another one for storage class and they are optional with detaults, but persistence is always there. I was actually hoping to add automated e2e tests for persistence...

let's try create a PV in the emphemeral cluster. According to testplatform team, we can create PV in our test cluster in the emphemeral clusters.
I think we can use some PV like this 
```
apiVersion: v1
kind: PersistentVolume
metadata:
  name: lightspeed-postgres-pv
spec:
  storageClassName: ""
  capacity:
    storage: 1Gi
  accessModes:
    - ReadWriteOnce
  local:
    path: /tmp
  nodeAffinity:
    required:
      nodeSelectorTerms:
        - matchExpressions:
            - key: kubernetes.io/os
              operator: In
              values:
                - linux
  claimRef:
    name: lightspeed-postgres-pvc
```
Its `claimRef` correspond to the PVC creatred in this PR.

internal/controller/ols_app_postgres_assets_test.go
```diff
@@ -131,6 +131,9 @@ var _ = Describe("App postgres server assets", func() {
 				Name: PostgresDataVolume,
 				VolumeSource: corev1.VolumeSource{
 					EmptyDir: &corev1.EmptyDirVolumeSource{},
+					//					PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
```
raptorsun, line 134: let's add a test case to generate the PVC,too.

syedriko: yep, this is WIP

syedriko:
@xrajesh @raptorsun PTAL. I described the persistent storage behavior in https://issues.redhat.com/browse/OLS-1754. By default, there's no persistence, it needs to be enabled explicitly.

internal/controller/olsconfig_controller.go
```diff
@@ -110,6 +110,10 @@ type OLSConfigReconcilerOptions struct {
 // NetworkPolicy for restricting access to OLS pods
 // +kubebuilder:rbac:groups=networking.k8s.io,namespace=openshift-lightspeed,resources=networkpolicies,verbs=get;list;watch;create;update;patch;delete
 
+// PVC access for the Postgres PVC
+// +kubebuilder:rbac:groups=core,namespace=openshift-lightspeed,resources=persistentvolumeclaims,verbs=get;list;watch;create;update;patch;delete
```
xrajesh, line 114: @syedriko operator can be installed in any namespace , we will not have openshift-lightspeed namespace at that time.

syedriko: @xrajesh PTAL. I removed `namespace=openshift-lightspeed` everywhere in that file and replaced `groups=core` with `groups=""` as it's more idiomatic.

raptorsun:
/override "Red Hat Konflux / service-e2e-tests-417 / ols-bundle"

raptorsun:
`Red Hat Konflux / service-e2e-tests-417 / ols-bundle` failed only tool calling cases

PR #879, opened 2025-07-25
[WIP] OLS-1435: add validation webhook for OLSConfig CRD

raptorsun:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing


raptorsun:
> @raptorsun did you look at using CRD schema validation rules to do this instead of a webhook? I'm pretty sure it's possible to define schema rules for most, if not all, of these checks and avoid running/managing the webhook (not to mention the webhook becomes a point of failure since the CR instances can't be managed if the webhook is down for some reason):
> 
> https://kubernetes.io/blog/2022/09/23/crd-validation-rules-beta/
> 
> even if you don't want to use the beta feature (CEL), there's also validation that can be done on things like field name (for the `cluster` name requirement): https://book.kubebuilder.io/reference/markers/crd-validation

Yes, I have checked the validation tools provided by kubebuilder. CRD validation rules works well when all the required information comes from the object itself. Last year we have investigated several config problems reported by users, many of which leads to a reference to inexistent secret, configmap. CRD validation rules cannot verify reffered resources so I'd like to put a validation webhook here to prevent users mistakenly change the OLSConfig with reference to inexistent resources. This PR starts with the token secret. In future we will add reference to configmaps such as those containing TLS certs, too.

raptorsun:
> > CRD validation rules cannot verify reffered resources so I'd like to put a validation webhook here to prevent users mistakenly change the OLSConfig with reference to inexistent resources. This PR starts with the token secret. In future we will add reference to configmaps such as those containing TLS certs, too.
> 
> If someone deletes the referenced configmap/secret, does the operator reconciliation flag that in status the next time it reconciles the olsconfig object?

yes, errors during reconciliation is put into the status of OLSConfig.

PR #997, opened 2025-09-22
Fixing OLS-1949

blublinsky:
## Description
This PR adds Deployment status checks to make sure that a deployment is not only created, but is also working correctly.

## Type of change

- [ ] Refactor
- [ x] New feature
- [ x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
[OLS-1949](https://issues.redhat.com//browse/OLS-1949)
- Closes #
[OLS-1949](https://issues.redhat.com//browse/OLS-1949)

- [x ] I have performed a self-review of my code.
- [ x] PR has passed all pre-merge test jobs.

## Testing
Creating a deployment with correct parameters works with no change 
Specifying a wrong model produces the following status:
```
Status:
  Conditions:
    Last Transition Time:  2025-09-22T08:36:31Z
    Message:               In Progress: failed to reconcile App Deployment: deployment not ready: 0 replicas available
    Reason:                Reconciling
    Status:                False
    Type:                  Reconciled
    Last Transition Time:  2025-09-22T08:36:32Z
    Message:               All components are successfully deployed
    Reason:                Reconciling
    Status:                True
    Type:                  ConsolePluginReady
    Last Transition Time:  2025-09-22T08:36:36Z
    Message:               All components are successfully deployed
    Reason:                Reconciling
    Status:                True
    Type:                  CacheReady
```
If the models in LLM and OLS do not match, we also get the correct status:
```
Status:
  Conditions:
    Last Transition Time:  2025-09-22T08:39:27Z
    Message:               In Progress: failed to reconcile App Deployment: deployment not ready: 0 replicas available
    Reason:                Reconciling
    Status:                False
    Type:                  Reconciled
    Last Transition Time:  2025-09-22T08:39:27Z
    Message:               All components are successfully deployed
    Reason:                Reconciling
    Status:                True
    Type:                  ConsolePluginReady
    Last Transition Time:  2025-09-22T08:39:31Z
    Message:               All components are successfully deployed
    Reason:                Reconciling
    Status:                True
    Type:                  CacheReady
```
    
See above

### DIFF
--- a/internal/controller/ols_app_postgres_reconciliator.go
+++ b/internal/controller/ols_app_postgres_reconciliator.go
@@ -15,7 +15,7 @@ import (
 	olsv1alpha1 "github.com/openshift/lightspeed-operator/api/v1alpha1"
 )
 
-func (r *OLSConfigReconciler) reconcilePostgresServer(ctx context.Context, olsconfig *olsv1alpha1.OLSConfig) error {
+func (r *OLSConfigReconciler) reconcilePostgresServer(ctx context.Context, olsconfig *olsv1alpha1.OLSConfig) (string, error) {
 	r.logger.Info("reconcilePostgresServer starts")
 	tasks := []ReconcileTask{
 		{
@@ -38,33 +38,34 @@ func (r *OLSConfigReconciler) reconcilePostgresServer(ctx context.Context, olsco
 			Name: "reconcile Postgres PVC",
 			Task: r.reconcilePostgresPVC,
 		},
-		{
-			Name: "reconcile Postgres Deployment",
-			Task: r.reconcilePostgresDeployment,
-		},
+
 		{
 			Name: "generate Postgres Network Policy",
 			Task: r.reconcilePostgresNetworkPolicy,
 		},
+		{
+			Name: "reconcile Postgres Deployment",
+			Task: r.reconcilePostgresDeployment,
+		},
 	}
 
 	for _, task := range tasks {
-		err := task.Task(ctx, olsconfig)
+		message, err := task.Task(ctx, olsconfig)
 		if err != nil {
 			r.logger.Error(err, "reconcilePostgresServer error", "task", task.Name)
-			return fmt.Errorf("failed to %s: %w", task.Name, err)
+			return message, fmt.Errorf("failed to %s: %w", task.Name, err)
 		}
 	}
 
 	r.logger.Info("reconcilePostgresServer completed")
 
-	return nil
+	return "", nil
 }
 
-func (r *OLSConfigReconciler) reconcilePostgresDeployment(ctx context.Context, cr *olsv1alpha1.OLSConfig) error {
+func (r *OLSConfigReconciler) reconcilePostgresDeployment(ctx context.Context, cr *olsv1alpha1.OLSConfig) (string, error) {
 	desiredDeployment, err := r.generatePostgresDeployment(cr)
 	if err != nil {
-		return fmt.Errorf("%s: %w", ErrGeneratePostgresDeployment, err)
+		return "Failed", fmt.Errorf("%s: %w", ErrGeneratePostgresDeployment, err)
 	}
 
 	existingDeployment := &appsv1.Deployment{}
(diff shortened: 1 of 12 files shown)


raptorsun:
the function `checkDeploymentStatus` is called for console and database deployment, we need to check the app server deployment status, too.

blublinsky:
> the `ReconcileFunc` 's new return value of `message` seems able to be deduced from the error message it returns. for example, we return empty messge for empty error, and a non-empty message when error is set. the message content relates to the error message. Could we skip returning the message besides the error and deduce the message content based on the error?

In the current implementation on every error we automatically populate message as failed. In reality we need to distinguish between "Fail" and "In progress". When deployment is created but not ready yet, its not Failed

thoraxe:
I think this handles more cases than my https://github.com/openshift/lightspeed-operator/pull/992 right? I should close 992?

internal/controller/ols_app_postgres_reconciliator.go
```diff
@@ -38,33 +38,34 @@ func (r *OLSConfigReconciler) reconcilePostgresServer(ctx context.Context, olsco
...
-		},
+
 		{
 			Name: "generate Postgres Network Policy",
 			Task: r.reconcilePostgresNetworkPolicy,
 		},
+		{
+			Name: "reconcile Postgres Deployment",
+			Task: r.reconcilePostgresDeployment,
+		},
```
xrajesh, line 49: @blublinsky Is there any significance to change reconcilePostgresDeployment after reconcilePostgresNetworkPolicy - I feel the original order is better.

internal/controller/ols_app_server_assets_test.go
```diff
@@ -1908,18 +1908,18 @@ func generateRandomSecret() (*corev1.Secret, error) {
...
+//		ObjectMeta: metav1.ObjectMeta{
+//			Name:      "test-configmap",
+//			Namespace: "openshift-lightspeed",
+//		},
+//		Data: map[string]string{
+//			"service-ca.crt": "random ca cert content",
+//		},
+//	}
+//	return &configMap, nil
+//}
```
xrajesh, line 1922: @blublinsky we can remove this function ?

internal/controller/ols_console_reconciliator.go
```diff
@@ -52,25 +49,29 @@ func (r *OLSConfigReconciler) reconcileConsoleUI(ctx context.Context, olsconfig
 			Name: "reconcile Console Plugin NetworkPolicy",
 			Task: r.reconcileConsoleNetworkPolicy,
 		},
+		{
+			Name: "reconcile Console Plugin Deployment",
+			Task: r.reconcileConsoleUIDeployment,
+		},
```
xrajesh, line 55: @blublinsky Similar question , why do we need to change the reconcile order ?

internal/controller/ols_app_server_assets_test.go
blublinsky: I think we should
It was there and I did not want to delete it straight,cause I was not the one putting it in

internal/controller/ols_console_reconciliator.go
blublinsky: similar answer as above

internal/controller/ols_app_postgres_reconciliator.go
```diff
@@ -81,146 +82,155 @@ func (r *OLSConfigReconciler) reconcilePostgresDeployment(ctx context.Context, c
 		r.logger.Info("creating a new OLS postgres deployment", "deployment", desiredDeployment.Name)
 		err = r.Create(ctx, desiredDeployment)
 		if err != nil {
-			return fmt.Errorf("%s: %w", ErrCreatePostgresDeployment, err)
+			return "Failed", fmt.Errorf("%s: %w", ErrCreatePostgresDeployment, err)
 		}
-		return nil
+
+		// Deployment was just created – it cannot be ready yet.
+		return "In Progress", fmt.Errorf("deployment is not ready, 0 replicas available")
```
raptorsun, line 89: I doubt a deployment progressing to running phase is considered an error. It is expected behavior. we can check the returned message and queue a follow-up reconciliation several seconds later in the main reconciliation loop.

Moreover, a deployment just get updated or created has no chance to achieve ready state during the operator function call that only lasts several miliseconds. 

I suggest we check all the deployments' statuses at the end of main reconciliation loop and set the deployment status accordingly.

blublinsky: We do not report this as an error. Its in Progress

raptorsun: it is returning an error, caught by the outer loop, it will interrupt the reconciliation.

raptorsun:
shall we delay the deployment ready check to the main reconciliation loop just before the status settings? https://github.com/openshift/lightspeed-operator/blob/main/internal/controller/olsconfig_controller.go#L186
If any deployment is not fully rolled out, we enqueue a quicker reconciliation to check the status.

PR #1430, opened 2026-03-19
OLS-2460 Wire TLS security profile to service config

onmete:
## Summary
- add `tlsSecurityProfile` to generated appserver `olsconfig.yaml` so service endpoint TLS follows operator/user CR configuration
- resolve profile via existing TLS helpers with Intermediate defaults when CR field is unset
- add appserver asset tests for default Intermediate, explicit Modern, and Custom profile passthrough

## Test plan
- [x] `make test` (appserver suite passes; command fails in this environment due local Go toolchain mismatch `go1.25.7` vs `go1.25.6` while running all internal packages)
- [x] Verified appserver test expectations updated for new `tlsSecurityProfile` field in generated config map

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/controller/appserver/assets.go
+++ b/internal/controller/appserver/assets.go
@@ -8,7 +8,10 @@ import (
 	"slices"
 	"strings"
 
+	configv1 "github.com/openshift/api/config/v1"
+
 	"github.com/openshift/lightspeed-operator/internal/controller/reconciler"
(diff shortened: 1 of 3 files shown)


onmete:
fips-or-die failure (exit 141 = SIGPIPE
/retest

PR #1431, opened 2026-03-19
OLS-2689 - Fixing upgrade tests

JoaoFula:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing


test/e2e/client.go
```diff
@@ -854,20 +856,107 @@ func (c *Client) CreateClusterRoleBinding(namespace, serviceAccount, clusterRole
...
 	}
 
-	err := cmd.Run()
+	if len(subscriptionList.Items) == 0 {
+		return fmt.Errorf("no subscriptions found in namespace %s", namespace)
+	}
+
+	subscription := &subscriptionList.Items[0]
+
+	currentCSV, found, err := unstructured.NestedString(subscription.Object, "spec", "name")
```
blublinsky: In OLM, spec.name is the catalog package name (e.g. lightspeed-operator), not a CSV. The values that actually describe CSV state are under status (e.g. installedCSV / currentCSV).

test/e2e/client.go
```diff
@@ -854,20 +856,107 @@ func (c *Client) CreateClusterRoleBinding(namespace, serviceAccount, clusterRole
...
+	if lightspeedPackageManifest == nil {
+		return fmt.Errorf("lightspeed-operator packagemanifest not found in openshift-marketplace")
+	}
+
+	channels, found, err := unstructured.NestedSlice(lightspeedPackageManifest.Object, "status", "channels")
+	if err != nil || !found || len(channels) == 0 {
+		return fmt.Errorf("failed to get channels from packagemanifest: %w", err)
+	}
+
+	firstChannel, ok := channels[0].(map[string]interface{})
```
blublinsky: The subscription already declares which catalog channel it uses: spec.channel. Taking channels[0] assumes the first entry in PackageManifest.status.channels is always that channel. If order or defaults differ, latestCSV can be for the wrong channel, so startingCSV may not match what you think you’re upgrading to.

Fix direction

After 878 (once you have subscription), read spec.channel (e.g. NestedString(..., "spec", "channel")), then in the loop over channels (you’d iterate 912’s slice instead of indexing [0]), pick the map whose name equals that channel, then read currentCSV from that map (same idea as 922).

.tekton/integration-tests/pipelines/lightspeed-operator-upgrade-e2e-test-pipeline-419.yaml
```diff
@@ -251,16 +195,31 @@ spec:
...
+              export OPERATOR_SDK_VERSION=1.36.1
+              export ARCH=$(case $(uname -m) in x86_64) echo -n amd64 ;; aarch64) echo -n arm64 ;; *) echo -n $(uname -m) ;; esac)
+              export OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/v${OPERATOR_SDK_VERSION}
+              curl -Lo /usr/local/bin/operator-sdk ${OPERATOR_SDK_DL_URL}/operator-sdk_linux_${ARCH}
+              chmod +x /usr/local/bin/operator-sdk
+              operator-sdk version
+              echo "---------------------------------------------"
+              oc create namespace $(params.namespace)
+              oc label namespaces $(params.namespace) openshift.io/cluster-monitoring=true --overwrite=true
+              LATEST_BUNDLE=$(ls lightspeed-catalog-4.19/bundle-v* | sort -V -r | head -n 2 | tail -n 1)
```
blublinsky: This seems fragile
Hidden assumption: There must be at least two bundle-v* files. If someone adds only one file, or the glob matches one file, head -n 2 | tail -n 1 can pick the only file (same as newest) or behave unexpectedly depending on ls output—so you might not be testing “upgrade from N−1” anymore.

test/e2e/client.go
```diff
@@ -854,20 +856,107 @@ func (c *Client) CreateClusterRoleBinding(namespace, serviceAccount, clusterRole
...
+	if err != nil {
+		return fmt.Errorf("failed to list subscriptions in namespace %s: %w", namespace, err)
 	}
 
-	err := cmd.Run()
+	if len(subscriptionList.Items) == 0 {
+		return fmt.Errorf("no subscriptions found in namespace %s", namespace)
+	}
+
+	subscription := &subscriptionList.Items[0]
```
blublinsky: Items[0] is whatever the API returns first for SubscriptionList in that namespace. If there is more than one Subscription (another operator, a leftover, a second install), you might update the wrong subscription’s startingCSV while thinking you’re driving the Lightspeed upgrade.

After listing, filter to the subscription you care about (e.g. spec.name == "lightspeed-operator" for the catalog package, and/or a known metadata.name), require exactly one match, then use that item instead of [0] blindly.

.tekton/integration-tests/pipelines/lightspeed-operator-upgrade-e2e-test-pipeline-419.yaml
JoaoFula: but if there's only one file then we shouldn't be testing the upgrade at all since it's pointless right?
I mean, I can add an exception where if only one file is found or grabbed, for us to fail the tests with "only one file found" but we can't really resume tests from that point either ways

test/e2e/client.go
JoaoFula: I'll make the change for sturdiness reasons and being correct but it won't really change much since the test environments are always sandboxes and the namespace doesn't exist until we create it. yes, it might affect one day that we decide to add a second operator installing in parallel but that's a long shot.

PR #1472, opened 2026-03-31
OLS-2670 - Adding multiple configuration test

JoaoFula:
## Description

Adding multiple configuration test to run periodically

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/test/e2e/all_features_test.go
+++ b/test/e2e/all_features_test.go
@@ -1,0 +1,1007 @@
+// all_features_test.go contains comprehensive end-to-end tests with ALL OLSConfig features enabled.
+// This test creates a single OLSConfig with all features turned on and runs individual validation
+// tests against that comprehensive configuration. It combines features from all other E2E tests plus
+// additional features not tested elsewhere (quotas, introspection, MCP servers, tool filtering, etc.).
+
+package e2e
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	olsv1alpha1 "github.com/openshift/lightspeed-operator/api/v1alpha1"
+	"github.com/openshift/lightspeed-operator/internal/controller/utils"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// All features test constants
+	allFeaturesPVCName             = "squid-volume-claim-all-features"
+	allFeaturesSquidDeploymentName = "squid-deployment-all-features"
+	allFeaturesSquidServiceName    = "squid-service-all-features"
+	allFeaturesSquidHTTPSPort      = 3349
+	allFeaturesSquidHTTPPort       = 3128
+	allFeaturesSquidConfigName     = "squid-config-all-features"
+	allFeaturesProxyConfigmapName  = "proxy-ca-all-features"
+	allFeaturesAdditionalCAName    = "additional-ca-certs-all-features"
+	allFeaturesMCPSecretName       = "mcp-auth-secret-all-features"
+	allFeaturesByokPullSecretName  = "byok-pull-secret-all-features"
+	allFeaturesImageName           = "assisted-installer-guide"
+	allFeaturesImageTag            = "latest"
+	allFeaturesInternalRegistry    = "image-registry.openshift-image-registry.svc:5000"
+	allFeaturesOrigImage           = "docker://quay.io/openshift-lightspeed-test/assisted-installer-guide:2025-1"
+)
+
+// Test Design Notes:
+// - Uses Ordered to ensure serial execution (all tests share single cluster-scoped CR)
+// - Tests ALL OLSConfig features in a single comprehensive configuration
+// - Labeled "AllFeatures" to exclude from standard make test-e2e runs
+// - Combines features from all other E2E tests plus new features not tested elsewhere
+// - FlakeAttempts(5) handles transient query timing and network issues
+// - Longer timeout (3h) due to complexity of setup and comprehensive testing
+var _ = Describe("All Features Enabled", Ordered, Label("AllFeatures"), FlakeAttempts(5), func() {
+	var env *OLSTestEnvironment
+	var err error
+	var client *Client
+	var cleanUpFuncs []func()
+	var storageClassName string
+	var squidHostname string
+	var registryDefaultRoute string
+	var builderToken string
+	var secret *corev1.Secret
+
+	BeforeAll(func() {
+		By("Setting up comprehensive all-features test environment")
+		ctx := context.Background()
+		client, err = GetClient(nil)
+		Expect(err).NotTo(HaveOccurred())
+
+		// Step 1: Detect or create storage class
+		By("Detecting or creating storage class")
+		storageClassName = StorageClassNameCI
+		defaultStorageClass, err := client.GetDefaultStorageClass()
+		if err == nil && defaultStorageClass != nil {
+			storageClassName = defaultStorageClass.Name
+		} else {
+			storageClassName = StorageClassNameLocal
+			By("Creating local storage class for testing")
+			cleanUpStorageClass, err := client.CreateStorageClass(storageClassName)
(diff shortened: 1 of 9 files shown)


raptorsun:
/lgtm
/approve

just a suggestion about the retry settings of flaky tests, maybe move the retry to individual tests will make the retry faster, if it is not neccessary to setup the environment again for the whole test suite.

test/e2e/all_features_test.go
```diff
@@ -0,0 +1,1009 @@
...
+			"llm_response": "OpenShift is a container platform",
+			"sentiment": 1
+		}`, conversationID)
+
+		feedbackResp, err := httpsClient.PostJson("/v1/feedback", feedbackReqBody, authHeader)
+		CheckEOFAndRestartPortForwarding(env, err)
+		Expect(err).NotTo(HaveOccurred())
+
+		// Exact status code may vary (400, 403, or 404) depending on implementation
+		Expect(feedbackResp.StatusCode).NotTo(Equal(http.StatusOK), "Feedback endpoint should be disabled")
```
blublinsky: I think this should be

Expect(feedbackResp.StatusCode).To(BeElementOf([]int{http.StatusBadRequest, http.StatusForbidden, http.StatusNotFound}), 
    "Feedback endpoint should return 4xx when disabled, not %d", feedbackResp.StatusCode)
Or if you wa

test/e2e/all_features_test.go
```diff
@@ -0,0 +1,1009 @@
...
+			for _, env := range container.Env {
+				if env.Name == "POSTGRES_SHARED_BUFFERS" || strings.Contains(env.Value, "512MB") {
+					foundSharedBuffers = true
+				}
+				if env.Name == "POSTGRES_MAX_CONNECTIONS" || strings.Contains(env.Value, "3000") {
+					foundMaxConnections = true
+				}
+			}
+		}
+		Expect(foundSharedBuffers || foundMaxConnections).To(BeTrue(), "Postgres configuration should include custom settings")
```
blublinsky: I think the above should be 
foundSharedBuffers := false
foundMaxConnections := false
for _, container := range postgresDeployment.Spec.Template.Spec.Containers {
    for _, env := range container.Env {
        if env.Name == "POSTGRES_SHARED_BUFFERS" && env.Value == "512MB" {
            foundSharedBuffers = true
        }
        if env.Name == "POSTGRES_MAX_CONNECTIONS" && env.Value == "3000" {
            foundMaxConnections = true
        }
    }
}
Expect(foundSharedBuffers).To(BeTrue(), "POSTGRES_SHARED_BUFFERS should be set to 512MB")
Expect(foundMaxConnections).To(BeTrue(), "POSTGRES_MAX_CONNECTIONS should be set to 3000")

test/e2e/all_features_test.go
```diff
@@ -0,0 +1,1009 @@
...
+		defaultStorageClass, err := client.GetDefaultStorageClass()
+		if err == nil && defaultStorageClass != nil {
+			storageClassName = defaultStorageClass.Name
+		} else {
+			storageClassName = StorageClassNameLocal
+			By("Creating local storage class for testing")
+			cleanUpStorageClass, err := client.CreateStorageClass(storageClassName)
+			Expect(err).NotTo(HaveOccurred())
+			cleanUpFuncs = append(cleanUpFuncs, cleanUpStorageClass)
+		}
```
blublinsky, line 82: cleanfunctions are not guaranteed to run on failure. use cleanUpFuncs,

test/e2e/all_features_test.go
```diff
@@ -0,0 +1,1009 @@
...
+			},
+		}
+		err = client.Get(appDeployment)
+		Expect(err).NotTo(HaveOccurred())
+
+		apiContainer := appDeployment.Spec.Template.Spec.Containers[0]
+		Expect(apiContainer.Resources.Limits).NotTo(BeNil())
+		Expect(apiContainer.Resources.Requests).NotTo(BeNil())
+		Expect(apiContainer.Resources.Limits.Cpu().String()).To(Equal("1"))
+		Expect(apiContainer.Resources.Limits.Memory().String()).To(Equal("2Gi"))
```
blublinsky, line 747: This assumes that API container is always the first one, which works, but is fragile in the case of changes
More reliable is to get it by name

var apiContainer *corev1.Container
for i := range appDeployment.Spec.Template.Spec.Containers {
    if appDeployment.Spec.Template.Spec.Containers[i].Name == ServerContainerName {
        apiContainer = &appDeployment.Spec.Template.Spec.Containers[i]
        break
    }
}

test/e2e/assets.go
```diff
@@ -52,90 +66,289 @@ func generateOLSConfig() (*olsv1alpha1.OLSConfig, error) { // nolint:unused
...
 				},
 			},
-		}, nil
+			CredentialsSecretRef: corev1.LocalObjectReference{
+				Name: LLMTokenSecondSecretName,
+			},
+			Type: llmProvider,
+		}
+		providers = append(providers, secondProvider)
+	}
```
blublinsky, line 117: Are we missing AzureDeploymentName and URL fields.

test/e2e/all_features_test.go
```diff
@@ -0,0 +1,1009 @@
...
+			if strings.Contains(line, `"query"`) && strings.Contains(line, "oldterm") && strings.Contains(line, `Body:`) {
+				// This is the raw HTTP request body, which is expected to contain oldterm
+				continue
+			}
+
+			// Look for the redacted query being processed
+			if strings.Contains(line, "newterm") && strings.Contains(line, "Kubernetes") {
+				foundRedactedQuery = true
+			}
+		}
```
blublinsky: This is very fragile
Assumes log format stability: The test skips lines containing Body: to avoid raw HTTP request logs. If OLS changes its logging format (e.g., changes Body: to Request: or Payload:), the test logic breaks.

Heuristic matching: Looking for lines with both "newterm" AND "Kubernetes" is a heuristic guess at what "processed query" logs look like. If log verbosity changes or the LLM processes the query differently, this could fail.

String counting: strings.Count(logs, "newterm") counts ALL occurrences in the entire log buffer - including debug output, error messages, repeated logs from retries, etc. The assumption "≥3 means filtering worked" could be invalidated by log noise.

No structured logging: The test parses free-form text instead of structured log entries (JSON logs with fields like event=query_filtered, original_term=oldterm, replacement=newterm).

test/e2e/all_features_test.go
```diff
@@ -0,0 +1,1009 @@
...
+		By("Verifying multiple 'newterm' replacements appear in logs")
+		newtermCount := strings.Count(logs, "newterm")
+		// We should see at least 3 occurrences of newterm (one for each oldterm)
+		// There might be more depending on how the logs are formatted
+		Expect(newtermCount).To(BeNumerically(">=", 3), "Should find at least 3 occurrences of 'newterm' in logs")
+
+		By("Verifying the query was processed successfully despite filtering")
+		bodyStr := string(body)
+		Expect(bodyStr).To(ContainSubstring("conversation_id"))
+	})
```
blublinsky, line 902: same as above. very fragile

blublinsky:
he test is labeled "AllFeatures" but some features (tools approval, introspection) are only verified via CR/ConfigMap checks, not actual behavioral tests. Is the intent to add behavioral coverage later, or is configuration-only validation sufficient?

JoaoFula:
This test is more to verify that with several (as much as possible) features enabled in the CR that OLS continues to work. We can later add behavioral tests later on if we deem it to be a necessity. Now that you added the MCP tests and we have introspection tests in service repo, we can indeed add them soon but not in scope for this PR Id say

test/e2e/all_features_test.go
```diff
@@ -0,0 +1,1009 @@
...
+		Expect(err).NotTo(HaveOccurred())
+		Expect(resp.StatusCode).To(Equal(http.StatusOK))
+		Expect(body).NotTo(BeEmpty())
+
+		By("Verifying response contains valid LLM output")
+		bodyStr := string(body)
+		Expect(bodyStr).To(ContainSubstring("conversation_id"))
+
+		By("Retrieving pod logs to verify query filter was applied")
+		logs, podName, err := GetAppServerPodLogs(client, &queryStartTime)
```
raptorsun: it may be useful to have a `Eventually().Should()`, for the case that log does not immediately appear after the query.

raptorsun:
looks good overall.
i only have suggestions on retrying the search on pod log to make it less probable to miss delayed log updates, as the log is async.

PR #1548, opened 2026-04-22
splitting e2e tests between bundle and operator

blublinsky:
## Description

**Summary**
Consolidates Konflux integration Tekton Pipeline YAML into one file per test family (operator e2e, operator upgrade e2e, service integration) with pipeline parameters for OpenShift minor line (openshift-version-prefix, artifact-oci-tag-prefix where needed, test-name, namespace). Removes 13 per-version pipeline files and the stale e2e-tests.yaml scenario that pointed at a non-existent pipeline.

Extracts bundle install logic into hack/install/ (install-operator-bundle.sh, _lib.sh) and reuses it from the merged pipelines and rapidast-scan.yaml (shallow clone at ols-bundle snapshot revision, then script).

Adds konflux-integration-test-scenarios.yaml with 12 IntegrationTestScenario resources (v1beta2) targeting the merged pipelines and passing the correct parameters per minor (4.16–4.19).

Upgrade pipeline: installs baseline bundle, then snapshot bundle using PRE_BUNDLE_IMAGE + BUNDLE_IMAGE and a single install-operator-bundle.sh invocation (matches script contract and make test-upgrade).

config/default/deployment-patch.yaml: intentionally unchanged (keeps __REPLACE_*__ placeholders; no pinned digests in this PR).

**Next steps (post-merge / follow-ups)**
Konflux operations

Remove or rename legacy IntegrationTestScenario objects if names conflict with the new CRs (operator-e2e-tests-*, etc.).
Update any required integration test or release policy references that still point at ols-e2e-tests-416 or the old pipeline path.
Confirm test-name semantics

Operator e2e and service integration scenarios both use ols-e2e-tests-4.xx as the test-name param while metadata.name differs. Verify Konflux/reporting does not require globally unique test-name values; adjust if needed.
Git resolver revision

Scenarios pin revision: main. If you want PRs to exercise pipeline YAML on the PR branch, adopt whatever revision substitution pattern your tenant supports (e.g. dynamic revision), or document that pipeline changes apply after merge to main.
Optional: IDMS parity

rapidast-scan.yaml sets IMAGE_DIGEST_MIRROR_SET_URL; merged operator e2e install does not. If ephemeral clusters show pull errors for registry.redhat.io content, export the same URL in the operator e2e install step.
Security / hygiene

Consider dropping cat $KUBECONFIG (or redacting) from Tekton step logs in a follow-up PR.
Direct install path (not in this PR)

install-operator-direct.sh exists but is not wired into a parallel Konflux pipeline or scenario.
Later: add a pipeline/scenario (or parameterized branch), SKIP_IDMS / Hypershift behavior as needed, and document when to use direct vs bundle.
Timeouts

If integration runs hit default timeouts, add Konflux IntegrationTestScenario annotations for pipeline/tasks/finally timeouts per [Konflux docs](https://konflux-ci.dev/docs/testing/integration/editing/).
Developer ergonomics

## Type of change

- [ ] Refactor
- [ x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
https://redhat.atlassian.net/browse/OLS-2857
- Closes #

## Testing


.tekton/integration-tests/pipelines/lightspeed-service-integration-test-pipeline-4.17.yaml
```diff
@@ -78,65 +104,129 @@ spec:
...
+              COMMIT="$(jq -r --arg component_name "${KONFLUX_COMPONENT_NAME}" '.components[] | select(.name == $component_name) | .source.git.revision' <<< "$SNAPSHOT")"
+              echo "Cloning lightspeed-operator@${COMMIT} for Konflux operator install"
+              rm -rf /workspace/lightspeed-operator
+              mkdir -p /workspace/lightspeed-operator
+              cd /workspace/lightspeed-operator
+              git init -q
+              git remote add origin https://github.com/openshift/lightspeed-operator.git
+              git fetch --depth 1 origin "${COMMIT}"
+              git checkout -q FETCH_HEAD
+              ./.tekton/integration-tests/scripts/run-konflux-operator-install.sh "$(params.install-mode)"
```
JoaoFula, line 180: Any reason for us to install the operator in the service tests considering we already have that logic in place in the service tests?

hack/install/install-operator-direct.sh
```diff
@@ -0,0 +1,132 @@
...
+	}
+	[[ -f "${REPO_ROOT}/hack/image_placeholders.json" ]] || {
+		echo "error: hack/image_placeholders.json not found" >&2
+		exit 1
+	}
+	[[ -f "${REPO_ROOT}/config/default/deployment-patch.yaml" ]] || {
+		echo "error: ${REPO_ROOT}/config/default/deployment-patch.yaml not found" >&2
+		exit 1
+	}
+	if [[ -z "$operator_img" ]]; then
```
JoaoFula: at script entry we're verifying that $IMG is not empty. Therefore this and the following ifs will never activate.

.tekton/integration-tests/scripts/run-konflux-operator-install.sh
```diff
@@ -0,0 +1,109 @@
...
+
+install_bundle() {
+	if [[ -n "${UPGRADE_E2E_INSTALL_OLD_BASE_FROM_CATALOG:-}" ]]; then
+		PV="${UPGRADE_E2E_INSTALL_OLD_BASE_FROM_CATALOG%.}"
+		catalog_dir="lightspeed-catalog-${PV}"
+		if [[ ! -d "${catalog_dir}" ]]; then
+			echo "error: catalog dir not found: ${catalog_dir} (cwd: $(pwd))" >&2
+			exit 1
+		fi
+		oldest_ver="$(ls "${catalog_dir}"/bundle-v*.yaml | sed -n 's/.*bundle-v\(.*\)\.yaml/\1/p' | sort -V | head -n1)"
```
JoaoFula: This will cause every upgrade test to fail since we're picking 1.0.8 and there's a known upgrade issue from 1.0.9 to 1.0.10. So if we pick 1.0.8 to any newer version, it will always fail. Thus why in my PR, I'm picking latest - 1. Assuming we're always testing it from now on, we never will run into problems and we will test all upgrades.

.tekton/integration-tests/pipelines/lightspeed-service-integration-test-pipeline-4.17.yaml
blublinsky: ols-install puts the operator from this SNAPSHOT on the cluster (bundle/direct via hack/install). The service step only runs lightspeed-service e2e (test-e2e-cluster.sh / pytest), which reconfigures OLS per suite (adapt_ols_config) and does not replace that snapshot install contract.

Pytest calls install_ols() only when the operator is not already there (OLM: CSV; direct: lightspeed-operator-controller-manager deployment). Otherwise it skips install and still runs adapt_ols_config() each run.

hack/install/install-operator-direct.sh
blublinsky: good catch, fixed

.tekton/integration-tests/scripts/run-konflux-operator-install.sh
blublinsky: Agreed — we switched the upgrade base from the oldest catalog entry to latest − 1 (second-newest semver in lightspeed-catalog-<minor>), with a hard error if there are fewer than two bundles. That avoids long broken chains from very old bases (e.g. 1.0.8) while still testing a real catalog → snapshot upgrade. Pipeline descriptions are updated to match.

PR #1582, opened 2026-04-30
Add agentic deploy scripts, CRDs, and operator integration

harche:
Adds agentic stack deploy scripts (`hack/agentic/`), agentic CRDs, `LightspeedAgents` feature gate, and operator integration for console/sandbox image management.


hack/agentic/lib.sh
```diff
@@ -0,0 +1,916 @@
...
+}
+
+###############################################################################
+# Full deploy helpers — shared by deploy.sh
+###############################################################################
+
+update_crds_and_rbac() {
+    step "Updating CRDs and RBAC"
+    cd "${OPERATOR_DIR}"
+    make manifests kustomize >/dev/null 2>&1
```
iNecas: Developer experience hint, many commands throw the output away, so the script fails without giving reasons why.

My agent suggests something like and I tent to agree

```bash
_run() {
    local _out
    _out=$(mktemp)
    if "$@" >"${_out}" 2>&1; then
        rm -f "${_out}"
    else
        local _rc=$?
        echo -e "    ${RED}✗${NC} Command failed: $*" >&2
        cat "${_out}" >&2
        rm -f "${_out}"
        return ${_rc}
    fi
}
```

hack/agentic/lib.sh
```diff
@@ -0,0 +1,916 @@
...
+  maxTurns: 100
+AGENTEOF
+        info "Agent CRs created (default, smart, fast)"
+    fi
+
+    ###########################################################################
+    # Step 3: AgenticConfig — Cluster Admin
+    # Approval policy, console/sandbox images, concurrency limits.
+    ###########################################################################
+    local AGENT_IMAGE="${INTERNAL_REG}/${NS_OPERATOR}/lightspeed-agentic-sandbox:${TAG}"
```
iNecas:
```suggestion
    local AGENT_IMAGE="${INTERNAL_REG}/${NS_OPERATOR}/${BC_AGENT}:${TAG}"
```

hack/agentic/lib.sh
```diff
@@ -0,0 +1,916 @@
...
+    info "Namespace ${NS_OPERATOR} ready"
+}
+
+deploy_operator_manifests() {
+    step "Deploying operator manifests"
+    local PATCH_FILE="config/default/deployment-patch.yaml"
+    cp "${PATCH_FILE}" "${PATCH_FILE}.bak"
+
+    local OPERATOR_IMG="${INTERNAL_REG}/${NS_OPERATOR}/lightspeed-operator:${TAG}"
+    local CONSOLE_IMG="${INTERNAL_REG}/${NS_OPERATOR}/lightspeed-console-plugin:${TAG}"
```
iNecas:
```suggestion
    local OPERATOR_IMG="${INTERNAL_REG}/${NS_OPERATOR}/${BC_OPERATOR}:${TAG}"
    local CONSOLE_IMG="${INTERNAL_REG}/${NS_OPERATOR}/${BC_CONSOLE}:${TAG}"
```

iNecas: Given all the other envs are also set in one place, might make sense to do this for `_IMG` vars as well. I see the `SKILLS_IMAGE` in redeploy.sh as well… the consistency could help a bit.

I know it's just a dev script, but still…

hack/agentic/lib.sh
```diff
@@ -0,0 +1,916 @@
...
+    local bc_name="$1" from_dir="$2" label="$3"
+    if [[ "${SKIP_BUILD}" == "true" ]]; then
+        warn "Skipping build of ${label}"
+        return 0
+    fi
+    oc patch "bc/${bc_name}" -n "${NS_OPERATOR}" --type=merge \
+        -p "{\"spec\":{\"output\":{\"to\":{\"name\":\"${bc_name}:${TAG}\"}}}}" >/dev/null 2>&1
+    echo "    Building ${label} on cluster (uploading source)..."
+    oc start-build "${bc_name}" -n "${NS_OPERATOR}" \
+        --from-dir="${from_dir}" --follow \
```
iNecas: needs `--wait` as well, otherwise it returns 0 even on failure.

hack/agentic/lib.sh
```diff
@@ -0,0 +1,916 @@
...
+      dockerfilePath: Containerfile
+EOF
+    info "BuildConfigs and ImageStreams ready"
+}
+
+# Build an image on the cluster via binary build.
+# Uploads the source directory to a builder pod, which runs the Dockerfile
+# natively on amd64 and pushes to the internal registry — no local container
+# engine, cross-compilation, registry route, or auth tokens needed.
+build_on_cluster() {
```
iNecas: given `build_on_cluster` and `start_build_async` are very similar, how about having just `_build sync` and `_build async` and handle the differences in particular places.

harche:
Great feedback @iNecas , thanks, I will update the PR and test with those changes.

hack/agentic/lib.sh
harche: Done — added `_run()` wrapper (lines 13-25) that captures stdout/stderr to a tempfile and only surfaces it on failure. All `oc apply`, `oc patch`, and `make` calls that previously redirected to `/dev/null` now go through `_run`.

hack/agentic/lib.sh
harche: Good catch — removed the standalone `AGENT_IMAGE` and switched to `AGENT_IMG` which is centralized in `lib.sh` (line 66), built from `${BC_AGENT}` + `${TAG}`. All four image vars are now defined once in `lib.sh`.

hack/agentic/lib.sh
harche: Done — `OPERATOR_IMG`, `CONSOLE_IMG`, `AGENT_IMG`, and `SKILLS_IMG` are now centralized globals in `lib.sh` (lines 64-67), all built from their respective `BC_*` constants. No more local definitions inside individual functions.

harche: Done — removed the local `SKILLS_IMAGE` from `redeploy-agent.sh` and switched to `SKILLS_IMG` from `lib.sh`. All four image vars are now defined once in `lib.sh` and shared across all scripts.

hack/agentic/lib.sh
harche: Done — added `--wait` alongside `--follow` in the unified `_build sync` path (line 234).

hack/agentic/lib.sh
harche: Done — merged `build_on_cluster` and `start_build_async` into a single `_build` function that takes `sync` or `async` as its first argument (lines 222-258), eliminating the duplicated patch/skip/build logic.

Makefile
```diff
@@ -64,7 +64,7 @@ OPERATOR_SDK_VERSION ?= v1.36.1
 IMG ?= $(IMAGE_TAG_BASE):$(VERSION)
 export IMG
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.27.1
+ENVTEST_K8S_VERSION = 1.35.0
```
harche, line 67: Bumped from 1.27.1 to match the K8s dependency version (k8s.io/api v0.35.4). The agentic CRDs use CEL functions like `format.dns1123Subdomain()` that were introduced in K8s 1.31+. Envtest 1.27 rejects these CRDs during test setup.

harche:
closing in favor of https://github.com/openshift/lightspeed-operator/pull/1641. 

/close

PR #1604, opened 2026-05-07
OLS-2459 propagate APIServer TLS profile to olsconfig

onmete:
## Summary

- When the OLSConfig CR does not specify a `tlsSecurityProfile`, the operator now reads the cluster-wide TLS profile from the `APIServer` CR (`config.openshift.io/v1`, name `cluster`) and writes it into the generated `olsconfig.yaml`. Falls back to Intermediate defaults if neither source has a profile.
- Adds an `APIServer` watch to `SetupWithManager` with a predicate that only triggers reconciliation when `Spec.TLSSecurityProfile` changes on the `cluster` object. The existing ConfigMap hash annotation mechanism handles the pod restart.
- Fixes a pre-existing Makefile typo where the ImageStream CRD copy was overwriting the APIServer CRD test fixture.

Companion to https://github.com/openshift/lightspeed-service/pull/2921 (service-side Postgres TLS hardening).

## Test plan

- [x] Existing `appserver` unit tests pass (106 specs, 79.2% coverage)
- [x] New test: `should generate configmap with TLS profile from APIServer when CR has none` — sets Modern profile on APIServer, leaves CR nil, asserts generated config uses Modern
- [x] All other test packages pass (`controller`, `console`, `lcore`, `postgres`, `utils`, `watchers`, `tls`)

Made with [Cursor](https://cursor.com)


internal/controller/appserver/assets_test.go
```diff
@@ -209,6 +210,30 @@ var _ = Describe("App server assets", func() {
...
+
+			var olsConfigMap map[string]interface{}
+			err = yaml.Unmarshal([]byte(cm.Data[utils.OLSConfigFilename]), &olsConfigMap)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(olsConfigMap).To(HaveKeyWithValue("ols_config", HaveKeyWithValue("tlsSecurityProfile", MatchKeys(Options(IgnoreExtras), Keys{
+				"type":          Equal("ModernType"),
+				"minTLSVersion": Equal(string(configv1.TLSProfiles[configv1.TLSProfileModernType].MinTLSVersion)),
+			}))))
+
+			apiServer.Spec.TLSSecurityProfile = nil
```
xrajesh: In case of failure above this may not run - may be we can use DeferCleanup func @onmete

onmete:
@xrajesh Good catch — moved the APIServer cleanup into `DeferCleanup` so it runs even if the assertions fail. Fixed in d5f3cf5.

PR #1726, opened 2026-06-16
OLS-3192 Add preStop lifecycle hook to postgres pod to prevent crash-loop afte…

sriroopar:
…r unclean shutdown

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing

## Release Notes

* **Bug Fixes**
  * Improved PostgreSQL shutdown by adding a graceful `preStop` stop command and explicitly enforcing the pod termination grace period to 90 seconds, aligned with the configured shutdown timeout.

* **Tests**
  * Expanded unit and end-to-end coverage to validate PostgreSQL shutdown lifecycle behavior (`preStop` exec command) and `terminationGracePeriodSeconds`, and strengthened comparisons for lifecycle hooks and termination grace period handling.

### DIFF
--- a/internal/controller/postgres/deployment.go
+++ b/internal/controller/postgres/deployment.go
@@ -2,6 +2,7 @@ package postgres
 
 import (
 	"context"
+	"fmt"
 	"path"
 	"strconv"
 
@@ -234,10 +235,21 @@ func GeneratePostgresDeployment(r reconciler.Reconciler, ctx context.Context, cr
 									Value: strconv.Itoa(cr.Spec.OLSConfig.ConversationCache.Postgres.MaxConnections),
 								},
 							},
+							Lifecycle: &corev1.Lifecycle{
+								PreStop: &corev1.LifecycleHandler{
+									Exec: &corev1.ExecAction{
+										Command: []string{
+											"/bin/sh", "-c",
+											fmt.Sprintf("if [ -f %s/PG_VERSION ]; then pg_ctl stop -D %s -m fast -w -t %d; fi", utils.PostgresUserDataDir, utils.PostgresUserDataDir, utils.PostgresShutdownTimeoutSeconds),
+										},
+									},
+								},
+							},
 						},
 					},
-					Volumes:            volumes,
-					ServiceAccountName: utils.PostgreServiceAccountName,
+					Volumes:                       volumes,
+					ServiceAccountName:            utils.PostgreServiceAccountName,
+					TerminationGracePeriodSeconds: &[]int64{utils.PostgresTerminationGracePeriodSeconds}[0],
 				},
 			},
 			RevisionHistoryLimit: &revisionHistoryLimit,
--- a/internal/controller/utils/constants.go
+++ b/internal/controller/utils/constants.go
@@ -417,6 +417,17 @@ ssl_ca_file = '/etc/certs/cm-olspostgresca/service-ca.crt'
 	AlertsAdapterAlertmanagerURLEnvVar = "ALERTMANAGER_URL"
 	// AlertsAdapterComponentLabel is the app.kubernetes.io/component label value for alerts adapter resources
 	AlertsAdapterComponentLabel = "alerts-adapter"
+	// PostgresTerminationGracePeriodSeconds is the grace period for postgres pod termination.
+	// This must be greater than PostgresShutdownTimeoutSeconds to allow pg_ctl to complete.
+	PostgresTerminationGracePeriodSeconds = int64(60)
+
+	// PostgresShutdownTimeoutSeconds is the timeout for pg_ctl stop command.
+	// This must be less than PostgresTerminationGracePeriodSeconds to complete before SIGKILL.
+	PostgresShutdownTimeoutSeconds = int64(55)
+
+	// PostgresUserDataDir is the PostgreSQL data directory path.
+	// PostgreSQL's initdb creates this structure: PGDATA/data/userdata
+	PostgresUserDataDir = PostgresDataVolumeMountPath + "/data/userdata"
 
 	// PostgresPVCName is the name of the PVC for the OLS Postgres server
 	PostgresPVCName = "lightspeed-postgres-pvc"
--- a/internal/controller/utils/utils.go
+++ b/internal/controller/utils/utils.go
@@ -204,7 +204,8 @@ func DeploymentSpecEqual(a, b *appsv1.DeploymentSpec, compareInitContainers bool
 		!apiequality.Semantic.DeepEqual(a.Strategy, b.Strategy) || // check strategy
 		!PodVolumeEqual(a.Template.Spec.Volumes, b.Template.Spec.Volumes) || // check volumes
 		*a.Replicas != *b.Replicas || // check replicas
-		a.Template.Spec.ServiceAccountName != b.Template.Spec.ServiceAccountName { // check service account name
+		a.Template.Spec.ServiceAccountName != b.Template.Spec.ServiceAccountName || // check service account name
+		!apiequality.Semantic.DeepEqual(a.Template.Spec.TerminationGracePeriodSeconds, b.Template.Spec.TerminationGracePeriodSeconds) { // check termination grace period
 		return false
 	}
 
@@ -250,7 +251,8 @@ func ContainerSpecEqual(a, b *corev1.Container) bool {
 		a.ImagePullPolicy == b.ImagePullPolicy && // check image pull policy
 		ProbeEqual(a.LivenessProbe, b.LivenessProbe) && // check liveness probe
 		ProbeEqual(a.ReadinessProbe, b.ReadinessProbe) && // check readiness probe
-		ProbeEqual(a.StartupProbe, b.StartupProbe)) // check startup probe
+		ProbeEqual(a.StartupProbe, b.StartupProbe) && // check startup probe
+		apiequality.Semantic.DeepEqual(a.Lifecycle, b.Lifecycle)) // check lifecycle hooks
 }
 
 // EnvEqual compares two EnvVar slices ignoring order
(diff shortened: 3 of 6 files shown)


internal/controller/postgres/deployment.go
```diff
@@ -234,10 +235,21 @@ func GeneratePostgresDeployment(r reconciler.Reconciler, ctx context.Context, cr
 									Value: strconv.Itoa(cr.Spec.OLSConfig.ConversationCache.Postgres.MaxConnections),
 								},
 							},
+							Lifecycle: &corev1.Lifecycle{
+								PreStop: &corev1.LifecycleHandler{
+									Exec: &corev1.ExecAction{
+										Command: []string{
+											"/bin/sh", "-c",
+											fmt.Sprintf("pg_ctl stop -D %s -m fast -w -t %d", utils.PostgresUserDataDir, utils.PostgresShutdownTimeoutSeconds),
+										},
```
sriroopar, line 244: **Allocation for a pointer to constant**: `&[]int64{utils.PostgresTerminationGracePeriodSeconds}[0]` allocates a slice just to get a `*int64`. This is a known Go pattern but it's fragile and non-obvious.

Consider a simple helper or inline variable:
```go
terminationGracePeriod := utils.PostgresTerminationGracePeriodSeconds
// ...
TerminationGracePeriodSeconds: &terminationGracePeriod,
```
Or use `k8s.io/utils/ptr.To(utils.PostgresTerminationGracePeriodSeconds)` which is already a transitive dependency.

internal/controller/utils/constants.go
```diff
@@ -330,6 +330,17 @@ ssl_ca_file = '/etc/certs/cm-olspostgresca/service-ca.crt'
...
 	// AlertsAdapterComponentLabel is the app.kubernetes.io/component label value for alerts adapter resources
 	AlertsAdapterComponentLabel = "alerts-adapter"
+	// PostgresTerminationGracePeriodSeconds is the grace period for postgres pod termination.
+	// This must be greater than PostgresShutdownTimeoutSeconds to allow pg_ctl to complete.
+	PostgresTerminationGracePeriodSeconds = int64(60)
+
+	// PostgresShutdownTimeoutSeconds is the timeout for pg_ctl stop command.
+	// This must be less than PostgresTerminationGracePeriodSeconds to complete before SIGKILL.
+	PostgresShutdownTimeoutSeconds = 55
+
```
sriroopar, line 427: **Type inconsistency**: `PostgresTerminationGracePeriodSeconds` is `int64(60)` but `PostgresShutdownTimeoutSeconds` is untyped `55` (defaults to `int`). These two constants have a mathematical relationship (grace period must exceed shutdown timeout), but the type mismatch means you can't directly compute the delta without a cast.

Consider making both `int64`, or at minimum adding a compile-time assertion:
```go
PostgresShutdownTimeoutSeconds = int64(55)
```

internal/controller/utils/constants.go
```diff
@@ -330,6 +330,17 @@ ssl_ca_file = '/etc/certs/cm-olspostgresca/service-ca.crt'
...
+	// PostgresUserDataDir is the PostgreSQL data directory path (relative to PostgresDataVolumeMountPath).
+	// PostgreSQL's initdb creates this structure: PGDATA/data/userdata
+	PostgresUserDataDir = "/var/lib/pgsql/data/userdata"
```
sriroopar: **Path duplication risk**: `PostgresUserDataDir = "/var/lib/pgsql/data/userdata"` hardcodes a path whose prefix is already defined as `PostgresDataVolumeMountPath = "/var/lib/pgsql"`. If someone changes the mount path, this constant won't track.

Consider deriving it:
```go
PostgresUserDataDir = PostgresDataVolumeMountPath + "/data/userdata"
```
Also — is `data/userdata` always the correct PGDATA subdirectory for the Red Hat postgres image? This path is an internal implementation detail of the container image's `initdb`. If the image ever changes its PGDATA layout, this will silently break (pg_ctl will report 'not a database directory' to stderr, but the preStop hook has no error propagation).

sriroopar (commented):
## Adversarial Review

The preStop hook approach is sound — `pg_ctl stop -m fast -w` is the right tool for clean Postgres shutdown in Kubernetes. Here are concerns to address:

### CI Note

The `ci/prow/bundle-e2e-4-21` failure is **unrelated** to this PR. It's a pre-existing flaky proxy test (`proxy_test.go:162`) — the Squid PVC isn't cleaned up between retries (`squid-volume-claim already exists`), causing all retry attempts to fail. This eventually hits the 2-hour Prow timeout. A `/retest` should pass.

internal/controller/postgres/deployment.go
```diff
@@ -234,10 +235,21 @@ func GeneratePostgresDeployment(r reconciler.Reconciler, ctx context.Context, cr
 									Value: strconv.Itoa(cr.Spec.OLSConfig.ConversationCache.Postgres.MaxConnections),
 								},
```
sriroopar, line 236: **Silent failure mode**: If `pg_ctl` is not in `$PATH` inside the container (different base image, stripped image, etc.), or if the data directory doesn't exist yet (fresh pod that crashed before initdb), this preStop hook fails silently. Kubernetes will still send SIGTERM after the hook exits with a non-zero code, but you lose the clean-shutdown guarantee.

Consider adding a guard:
```sh
/bin/sh -c 'command -v pg_ctl >/dev/null 2>&1 && pg_ctl stop -D ... -m fast -w -t 55 || true'
```
The `|| true` ensures the hook doesn't block termination if pg_ctl isn't available.

internal/controller/utils/constants.go
```diff
@@ -330,6 +330,17 @@ ssl_ca_file = '/etc/certs/cm-olspostgresca/service-ca.crt'
 	AlertsAdapterAlertmanagerURLEnvVar = "ALERTMANAGER_URL"
 	// AlertsAdapterComponentLabel is the app.kubernetes.io/component label value for alerts adapter resources
 	AlertsAdapterComponentLabel = "alerts-adapter"
+	// PostgresTerminationGracePeriodSeconds is the grace period for postgres pod termination.
+	// This must be greater than PostgresShutdownTimeoutSeconds to allow pg_ctl to complete.
```
sriroopar, line 421: **5-second margin may be tight**: The 5-second gap between `PostgresShutdownTimeoutSeconds` (55) and `PostgresTerminationGracePeriodSeconds` (60) needs to cover:
1. Kubelet receives termination signal
2. Kubelet invokes the preStop hook shell (`/bin/sh -c ...`)
3. pg_ctl starts and begins shutdown

Steps 1-3 happen *before* the 55-second pg_ctl timer starts. If these steps take >5 seconds (which can happen under node pressure), pg_ctl will be SIGKILLed mid-shutdown — exactly the crash-loop scenario this PR aims to prevent.

Consider either increasing the grace period to 90s (with 55s pg_ctl timeout) or reducing the pg_ctl timeout to 45s.

internal/controller/utils/utils.go
```diff
@@ -194,7 +194,8 @@ func DeploymentSpecEqual(a, b *appsv1.DeploymentSpec, compareInitContainers bool
 		!apiequality.Semantic.DeepEqual(a.Strategy, b.Strategy) || // check strategy
 		!PodVolumeEqual(a.Template.Spec.Volumes, b.Template.Spec.Volumes) || // check volumes
 		*a.Replicas != *b.Replicas || // check replicas
-		a.Template.Spec.ServiceAccountName != b.Template.Spec.ServiceAccountName { // check service account name
+		a.Template.Spec.ServiceAccountName != b.Template.Spec.ServiceAccountName || // check service account name
+		!apiequality.Semantic.DeepEqual(a.Template.Spec.TerminationGracePeriodSeconds, b.Template.Spec.TerminationGracePeriodSeconds) { // check termination grace period
```
sriroopar, line 208: **Consistency note**: The existing `*a.Replicas != *b.Replicas` on line 196 will panic if either Replicas pointer is nil. The new `TerminationGracePeriodSeconds` comparison correctly uses `apiequality.Semantic.DeepEqual` which is nil-safe. This is good — but worth noting the pre-existing nil-unsafety on Replicas as a follow-up.

blublinsky (changes requested):
## Review — OLS-3192 preStop lifecycle hook

**Summary:** Adds a `preStop` hook running `pg_ctl stop -m fast` (guarded by `PG_VERSION` existence check) plus `terminationGracePeriodSeconds=60` to prevent CrashLoopBackOff after unclean postgres shutdown. Also extends `DeploymentSpecEqual` / `ContainerSpecEqual` for lifecycle + TGPS, and defaults nil TGPS → 30 in `SetDefaults_Deployment` to prevent spurious rollouts.

Overall the fix is solid and well-tested in unit tests. Two items to address:

---

### Must-fix: E2E preStop command doesn't match production

The unit test (`assets_test.go`) correctly expects the `if [ -f .../PG_VERSION ]` guard, but `test/e2e/all_features_test.go` asserts the **bare** `pg_ctl stop` command without the guard:

**Generator produces:**
```
if [ -f /var/lib/pgsql/data/userdata/PG_VERSION ]; then pg_ctl stop -D /var/lib/pgsql/data/userdata -m fast -w -t 55; fi
```

**E2E expects:**
```
pg_ctl stop -D /var/lib/pgsql/data/userdata -m fast -w -t 55
```

The `Equal` matcher will fail when AllFeatures e2e runs. CI is green because `bundle-e2e-4-21` excludes the `AllFeatures` label, but this will break when the full suite runs.

**Fix:** Sync the e2e expected string to include the `if [ -f …/PG_VERSION ]; then …; fi` wrapper (matching `assets_test.go`).

---

### Should-fix: Derive PostgresUserDataDir from mount path

`PostgresUserDataDir = "/var/lib/pgsql/data/userdata"` hardcodes a prefix that already exists as `PostgresDataVolumeMountPath = "/var/lib/pgsql"`. If someone changes the mount path, the userdata dir won't follow.

**Fix:** Derive it, e.g.:
```go
PostgresUserDataDir = PostgresDataVolumeMountPath + "/data/userdata"
```

---

### What's good

- Fix matches the Jira design and the known RHEL postgres unclean-shutdown failure mode
- `SetDefaults_Deployment` TGPS=30 is the correct companion — prevents spurious rollouts without it
- Lifecycle + TGPS equality coverage in comparison tests is thorough

internal/controller/utils/constants.go
```diff
@@ -417,6 +417,17 @@ ssl_ca_file = '/etc/certs/cm-olspostgresca/service-ca.crt'
...
+	PostgresTerminationGracePeriodSeconds = int64(60)
```
raptorsun, line 422: the value in the Jira ticket is longer ( 90 seconds ).
I dont think it blocking. Is this a result of testing optimal settings?

sriroopar: yes, this is based on test settings.

PR #1758, opened 2026-06-25
OLS-3349: Reconcile agentic console plugin in lightspeed-operator

blublinsky:
## Description

**Summary**
Migrate agentic console plugin reconciliation from lightspeed-agentic-operator into lightspeed-operator ([OLS-3236](https://redhat.atlassian.net/browse/OLS-3236), operator scope).

36 files — 9 added, 27 modified (~3.1k / ~710 lines).

**Operator reconciliation**
Add internal/controller/agenticconsole/ — Phase 1 (ConfigMap, NetworkPolicy, ServiceAccount) and Phase 2 (Deployment, Service, TLS wait, ConsolePlugin, Console activation), plus RemoveAgenticConsole() finalizer cleanup.
Wire into OLSConfigReconciler: Phase 1/2 steps, AgenticConsolePluginReady, TLS secret watcher, deployment restart mapping.
Add spec.ols.deployment.agenticConsole (Config) to OLSConfig; regenerate config/crd/bases/ols.openshift.io_olsconfigs.yaml.
Add --agentic-console-image in cmd/main.go (default: Konflux lightspeed-agentic-console:main).
Reuse existing resource names for upgrade adoption (lightspeed-agentic-console-plugin, cert lightspeed-agentic-console-plugin-cert, port 9443, container console).
No ConsolePlugin proxy to app-server (unlike chat console).
Shared console plugin refactor
Extract internal/controller/utils/console_plugin_reconciler.go (+ tests).
Refactor internal/controller/console/ to use shared helpers (~400 lines removed).

**Local dev fix**
Skip reconcileMetricsReaderSecret when LOCAL_DEV_MODE=true (make run) to stop metrics-reader-token reconcile loop. Add agentic console name to make run

**Documentation**
AGENTS.md, ARCHITECTURE.md
.ai/spec/what/agentic-console-ui.md (new), console-ui.md, README.md, how/project-structure.md, how/reconciliation.md

**Intentionally out of scope (follow-up)**
OLM bundle/CSV sync (bundle/manifests/, related_images.json, CSV --agentic-console-image)

## Type of change

- [ ] Refactor
- [ x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up dependent library

## Related Tickets & Documents

- Related Issue 
https://redhat.atlassian.net/browse/OLS-3349
- Closes #

## Checklist before requesting a review

[...]

PR #1779, opened 2026-07-02
OLS-3450: Add credentialHotReload flag to skip app-server restart on LLM secret rotation

lalan7:
## Summary

- Adds opt-in `credentialHotReload` boolean to OLSSpec CRD (default: `false`)
- When enabled, the operator's secret watcher skips rolling restart for LLM credential secrets — the service re-reads credentials from disk on each request ([RFE-9380](https://redhat.atlassian.net/browse/RFE-9380))
- Preserves existing behavior for all clusters when disabled

## Test plan

- [x] Unit tests: 3 new tests covering skip/no-skip/non-LLM paths (23/23 pass)
- [x] E2e test: rotate LLM secret with flag enabled, verify no restart via `Consistently`
- [x] `go vet ./...` clean
- [x] `go build ./...` clean
- [x] CRD regenerated via `make manifests`

## Related

- Jira: [OLS-3450](https://redhat.atlassian.net/browse/OLS-3450)
- RFE: [RFE-9380](https://redhat.atlassian.net/browse/RFE-9380)
- Service companion PR: openshift/lightspeed-service#2955

## Summary by CodeRabbit

- **New Features**
  - Added `spec.ols.credentialHotReload` to reload rotated LLM credentials without app-server restarts.
  - When enabled, supported services automatically read updated credential data.
- **Documentation**
  - Added guidance on prerequisites and behavior for credential data and secret reference changes.
- **Tests**
  - Added coverage verifying hot-reload configuration and watcher behavior when enabled or disabled.


lalan7:
Hi @blublinsky — this is the operator companion to service PR openshift/lightspeed-service#2955 ([RFE-9380](https://redhat.atlassian.net/browse/RFE-9380)).

Adds an opt-in `credentialHotReload` boolean to the OLSConfig CRD. When enabled, the operator skips rolling restarts for LLM credential secret rotations — the service handles it in-process. Default is `false`, so no behavior change for existing clusters.

All CI checks pass. The `bundle-e2e-4-21` failure is a pre-existing `main` issue (unknown fields `agenticConsole`/`alertsAdapter` in test fixtures from PRs #1758/#1759), not related to this change.

Would appreciate a review when you have a chance.

blublinsky:
**Suggestion: drive hot-reload from `ForEachExternalSecret` instead of `SecretWatcherFilter`**

The flag is currently wired in two places: `LLMSecretNames` + `CredentialHotReload` are populated before `ForEachExternalSecret`, and `SecretWatcherFilter` has a runtime skip branch. That duplicates the provider iteration and adds special-case logic on the event path.

A simpler approach is to handle this **inside the `ForEachExternalSecret` callback**, the same way TLS secrets get their `AnnotatedSecretMapping` today: when `credentialHotReload` is enabled and `source` is `llm-provider-*`, **do not annotate** that secret (skip `annotateSecretIfNeeded`). Without the watcher annotation, `shouldWatchSecret` never admits it, so **no restart watcher is created for that secret** — no `.data` events on the hot-reload path, no `SecretWatcherFilter` changes, no `LLMSecretNames` map, no extra fields on `WatcherConfig`, no skip logic at event time.

When the flag is off, behavior stays as today: foreach annotates LLM credential secrets and rotations trigger a rolling restart as they do now.

That keeps “which secrets does the operator react to?” entirely in the existing foreach/mapping configuration model, instead of splitting it between reconcile-time config and filter-time exceptions.

One thing to nail when toggling the flag on for clusters that already have annotated LLM secrets: reconcile may need to **remove** the watcher annotation when hot-reload is enabled, not only skip adding it for new ones.

---

I’d hold off on deeper review until this is reworked — the foreach-based approach removes most of the PR’s watcher-layer changes and is the pattern we already use for other secret types.

blublinsky:
**Follow-up: `credentialHotReload` must reach the service via OLS config for end-to-end behavior**

Today this flag lives only on the CR and in operator `WatcherConfig` (skip restart). It is **not** propagated into the app-server `olsconfig` ConfigMap (`buildOLSConfig` / `AppSrvConfigFile`) and the service PR has no corresponding config field either.

That leaves two independent behaviors with no runtime coupling:

| Layer | What happens |
|-------|----------------|
| Operator (`credentialHotReload: true`) | Skips pod restart on LLM secret rotation |
| Service (lightspeed-service #2955) | Re-reads `credentials_path` on every request — but unconditionally, not driven by this flag |

For this to work **end-to-end**, the flag should flow:

```
OLSConfig CR (spec.ols.credentialHotReload)
  → operator builds olsconfig ConfigMap
  → service reads config and enables hot-reload credential reads when true
```

Without that, enabling the CR flag on a cluster with an older service image (or before #2955 is deployed) silently disables restarts while credentials stay stale at startup — the CR comment warns about this, but nothing in config enforces it.

Suggest:

1. Add the field to the generated OLS config (`utils.OLSConfig` / `buildOLSConfig`).
2. Service side: gate `get_credentials()` disk re-read on that config flag (default `false` for backward compatibility).
3. Operator watcher skip (ideally via `ForEachExternalSecret` / no annotation) should follow the same CR flag.

That gives one user-facing switch that coordinates both “don’t restart” and “do re-read from disk,” instead of two repos assuming the other half is always correct.

blublinsky:
**Scoping: operator-only PR + bundle follow-up**

This PR is correctly scoped to the **operator only** — service changes stay in openshift/lightspeed-service#2955. Keep it that way.

Once the operator changes here are in good shape, suggest landing this PR without blocking on the OLM bundle, then a **follow-up PR** that runs `make manifests && make bundle` (or `hack/update_bundle.sh`) so `bundle/manifests/ols.openshift.io_olsconfigs.yaml` picks up `credentialHotReload`. Right now the field is only in `config/crd/bases/`; OLM installs use the bundle CRD, so the feature isn’t usable from the shipped bundle until that follow-up.

Order I’d expect:

1. Operator PR (this repo) — CRD + reconcile/watcher wiring + config pass-through to `olsconfig` once agreed
2. Service PR — gated `get_credentials()` re-read on config flag
3. Bundle follow-up — regenerate committed `bundle/` manifests

That keeps review focused and matches how we usually ship CRD schema vs bundle image updates.

lalan7:
@blublinsky 

 Addressed all code review findings in the latest push:

Service (lightspeed-service):

Consolidated AWS credential reads — Bedrock.load() now calls get_aws_credentials() once and passes the tuple through to _build_boto3_session() / _build_sigv4_auth() via an optional parameter. This cuts disk reads per load() from 6 to 3 and eliminates a potential mid-request partial rotation race (e.g. new access_key + old secret_key if rotation lands between calls).

Guarded optional role_arn read — get_aws_credentials() now checks os.path.isfile() before attempting to read the role_arn file, avoiding spurious warning logs when it's not configured (common scenario).

Operator (lightspeed-operator):

Collected removeSecretAnnotationIfNeeded errors — failures are now appended to the errs slice, consistent with annotateSecretIfNeeded error handling. Previously these were logged but silently swallowed.
Replaced hardcoded annotation key in e2e test — all 3 instances of "ols.openshift.io/watcher" now use utils.WatcherAnnotationKey.

Added unit tests for removeSecretAnnotationIfNeeded — covers annotation present (removed), annotation absent (no-op), and secret not found (no-op).

All service tests pass (28/28). Operator compiles and passes go vet cleanly (envtest unit tests require kubebuilder binaries available only in CI).

raptorsun:
we are going to have a discussion reviewing the design of this feature and then decide how to proceed.

lalan7:
@raptorsun — thanks for the heads up. Happy to join the design discussion or provide any additional context on the approach. Just to summarize where things stand:

I implemented all of @blublinsky's feedback (annotation-based approach via ForEachExternalSecret, ConfigMap propagation to the service, gated get_credentials() on the flag)
All Prow CI checks are green on both PRs ([operator #1779](https://github.com/openshift/lightspeed-operator/pull/1779) + [service #2955](https://github.com/openshift/lightspeed-service/pull/2955))
The feature is opt-in (credentialHotReload: false by default), no behavior change for existing clusters
Please let me know when the discussion is scheduled — I'd like to be there to answer any questions. Thanks!

blublinsky:
**should-fix: "How It Works" section describes the old watcher-filter approach, not the current annotation-based implementation**

`docs/credential-hot-reload.md` lines 44–52 still describe the original approach (commit 1) where the flag was stored in `WatcherConfig` and the watcher filter had skip logic. The reworked approach (commit 2) is entirely different — the operator removes the watcher annotation from LLM secrets, so the watcher predicate never admits the event.

Suggested rewrite for "How It Works":

1. During reconciliation, if `credentialHotReload` is enabled and the secret source is `llm-provider-*`, the operator **removes** the watcher annotation (or skips adding it for new secrets).
2. Without the annotation, the watcher predicate (`shouldWatchSecret`) does not admit the event — no restart happens.
3. When the flag is toggled off, the annotation is re-added on the next reconciliation, restoring restart behavior.
4. Non-LLM secrets (TLS, MCP, Postgres) are annotated and trigger restarts regardless of the flag.

blublinsky:
**should-fix: Squash commits before merge**

The PR has 6 commits including incremental review fixes and a revert (commit 5 is reverted by commit 6). Per repo conventions, please squash to a single logical commit before merge.

blublinsky:
**nice-to-have: Hot-reload info log fires on every reconciliation**

The `r.Logger.Info("credentialHotReload is enabled …")` in `olsconfig_controller.go` will log on every reconcile when the flag is enabled. Reconciliations can be frequent (unrelated secret/configmap changes, periodic resyncs). Consider logging only on transition (e.g. when the flag value changes from the previous reconcile) or demoting to `V(1)` debug level.

blublinsky:
Overall the implementation is clean — the annotation-based approach (removing the watcher annotation instead of adding filter-time exceptions) is the right design. The three comments above are the only findings. Please address the two should-fixes and we are happy to approve.

blublinsky:
**Summary of remaining issues:**

**✅ Fixed (blublinsky's should-fixes):**
- Commits squashed to 1
- Docs "How It Works" section updated with annotation-removal approach
- Info log demoted to debug level (`V(1)`)

**✅ Fixed (CodeRabbit major issues):**
- Concurrency: Added NOTE explaining MaxConcurrentReconciles=1 assumption + future mutex guidance
- Error handling: Errors from `removeSecretAnnotationIfNeeded` appended to `errs`
- Flaky test: Replaced `time.Sleep` with `Eventually` + 30s timeout

**❌ NOT FIXED (CodeRabbit):**
- **Logger pattern** (2 calls in `olsconfig_helpers.go`): Still using `r.Logger.Error` instead of `r.GetLogger().Error`
  - Lines with "Failed to remove annotation from secret" and "Failed to annotate secret"

**❓ Optional (CodeRabbit nitpick):**
- Test boilerplate: Extract repeated `createTestReconciler` setup into helper in `watchers_test.go`

**Action needed:** Fix the logger pattern (`r.Logger.Error` → `r.GetLogger().Error`) before merge. The test boilerplate is a nice-to-have.

xrajesh:
@lalan7 -Thank you .  Not specific to this PR - I am trying to understand this hot reload -  is for apitoken - is this a internal requirement that your org has where the token is valid only for a specific interval? If you are using Azure Entra Id , the access token is auto refreshed right ? Is this  hot reload of api token is better than entra_id auto token refresh. Thank you.

blublinsky:
**should-fix: protect shared `WatcherConfig` maps with a mutex**

The safety rationale in `internal/controller/utils/types.go` is not valid: `MaxConcurrentReconciles=1` serializes reconcile executions, but it does not guarantee that informer event handlers/predicates cannot read `WatcherConfig` while reconciliation is rebuilding it.

`annotateExternalResources()` replaces and updates `AnnotatedSecretMapping` / `AnnotatedConfigMapMapping` (`internal/controller/olsconfig_helpers.go:402-407, 418-421, 449-451`), while watcher callbacks read those maps (`internal/controller/watchers/watchers.go:283` and the corresponding ConfigMap path). That can result in a Go data race or concurrent map read/write panic.

Please protect this shared state with synchronization before relying on it for hot-reload behavior. A focused fix would be to add an `RWMutex` to `WatcherConfig`, expose small read/update methods, build replacement mappings locally during reconciliation, and publish each complete map under the write lock. Watcher handlers should use the read-locked accessors. The current comment should be updated rather than treating the access as safe because of the controller work queue.

blublinsky:
**nice-to-have: add behavioral coverage for credential rotation**

The new E2E test verifies the operator wiring (removing/restoring the watcher annotation and propagating `credential_hot_reload`), but it does not verify the user-visible behavior:

- updating LLM Secret data while hot reload is enabled does not restart the app-server;
- disabling hot reload causes a later LLM Secret update to restart it again;
- the service actually reads the rotated credential.

This is non-blocking for the operator PR because the service-side credential reload belongs in the companion service tests. Still, adding operator coverage for the restart/no-restart behavior—or documenting it as a follow-up across the two repositories—would make the feature safer to maintain.

blublinsky:
**recommendation: regenerate generated CRD artifacts instead of manually editing them**

The bundle CRD diff includes unrelated generated churn, including removal of `creationTimestamp`/`status` fields and addition of a document separator. Please regenerate the CRD/bundle artifacts using the repository's standard generation command (for example, `make manifests && make bundle` or `hack/update_bundle.sh`) rather than manually adjusting generated files.

That keeps `config/crd/bases/` and `bundle/manifests/` consistent with the Go API definition and avoids generator drift. The intended `credentialHotReload` schema change should be the only relevant generated difference.

lalan7:
> @lalan7 -Thank you . Not specific to this PR - I am trying to understand this hot reload - is for apitoken - is this a internal requirement that your org has where the token is valid only for a specific interval? If you are using Azure Entra Id , the access token is auto refreshed right ? Is this hot reload of api token is better than entra_id auto token refresh. Thank you.

@xrajesh Thanks for the question! This hot-reload is for file-based API key rotation — when an external process (CronJob, Vault, etc.) updates the Kubernetes Secret referenced by credentialsSecretRef. Without it, each secret update triggers a pod restart. With it, the service just re-reads the file — no restart needed.

Azure Entra ID token refresh is handled automatically by the SDK, so this feature isn't needed for that flow. They solve different problems.

blublinsky:
Please regenerate the generated manifests using `make manifests` and `make bundle`.

PR #1811, opened 2026-07-10
Bundle Update from Integration Test Snapshot

raptorsun:
This PR updates the related_images and bundle based on the latest snapshot from Konflux integration test.

## Summary by CodeRabbit

* **Chores**
  * Updated the release manifest metadata timestamp.
  * Refreshed the Lightspeed operator controller to a newer build, including its related image reference, so the packaged controller uses the latest image.


raptorsun:
bringing back release
this temporary fix is no longer needed

PR #1857, opened 2026-07-22
OLS-3574,OLS-3575,OLS-3576,OLS-3577: Deploy OpenShift MCP as standalone HTTPS service

blublinsky:
…ne HTTPS service

## Description

**Summary**
- Move the built-in OpenShift MCP server from an app-server sidecar to a standalone HTTPS operand (`internal/controller/ocpmcp/`), gated by `spec.ols.introspectionEnabled` (epic [OLS-3573](https://redhat.atlassian.net/browse/OLS-3573) / feature [OLS-3526](https://redhat.atlassian.net/browse/OLS-3526)).
- [OLS-3574](https://redhat.atlassian.net/browse/OLS-3574): `spec.ols.deployment.mcpServer` is `Config` (replicas, resources, tolerations, nodeSelector); MCP can scale.
- [OLS-3575](https://redhat.atlassian.net/browse/OLS-3575): Phase 1 reconciles ConfigMap (TOML), CA ConfigMap (`inject-cabundle`), ServiceAccount, and NetworkPolicy; when introspection is off, `ocpmcp.Remove()` tears down managed resources (including TLS Secret). Phase 2 reconciles Service (service-ca serving cert), waits for TLS, then Deployment; `MCPServerReady=True` / `Reason=NotConfigured` when disabled. TLS Secret watch is static in SystemResources and gated by `OpenShiftMCPServerTLSWatchEnabled`.
- [OLS-3576](https://redhat.atlassian.net/browse/OLS-3576): App-server sidecar removed; talks to `https://openshift-mcp-server.<ns>.svc:8443/mcp`, mounts MCP CA into `extra_ca`, and tracks CA hash only while introspection is enabled.
- [OLS-3577](https://redhat.atlassian.net/browse/OLS-3577): E2E expectations updated for standalone MCP Deployment/Service (no sidecar).
- **OTEL fix (included):** client and app-server collector trust now use the OpenShift service-ca bundle (`openshift-service-ca.crt` / `OLSCAConfigMap`) instead of the collector leaf `tls.crt`, which breaks gRPC/OpenSSL verify when mounted as a CA.
- Bundle/CSV/`related_images` updates are out of scope for this cutover PR.

## Type of change

- [ ] Refactor
- [x ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
https://redhat.atlassian.net/browse/OLS-3574, https://redhat.atlassian.net/browse/OLS-3575, https://redhat.atlassian.net/browse/OLS-3576 and https://redhat.atlassian.net/browse/OLS-3577
- Closes #
https://redhat.atlassian.net/browse/OLS-3574, https://redhat.atlassian.net/browse/OLS-3575, https://redhat.atlassian.net/browse/OLS-3576 and https://redhat.atlassian.net/browse/OLS-3577

## Testing

## Summary by CodeRabbit

* **New Features**
  * OpenShift MCP Server now runs as a standalone HTTPS Deployment and Service when introspection is enabled.
  * Added configurable MCP server deployment settings with CA injection, `MCPServerReady` status, and certificate-driven restarts.

* **Bug Fixes**
  * Improved OpenTelemetry and MCP TLS trust to consistently use the OpenShift service CA bundle.
  * Refined update detection so app-server redeploys when the MCP CA bundle changes (and clears hashes when disabled).

* **Documentation**
  * Updated configuration, CRD guidance, reconciliation flow, TLS, audit logging, and project structure for the standalone MCP architecture (including revised MCP config/TOML and endpoint examples).


vimalk78 (changes requested):
## Adversarial Review — Score: 62/100

**Mode:** Jira (OLS-3574, OLS-3575, OLS-3576, OLS-3577) | **Round:** 1

### Adherence: 17/20 AC passed, 3 failed

The core reconciliation, TLS setup, CRD change, sidecar removal, and app-server wiring are all solid. The three failures are all in OLS-3577 (E2E):

#### FAIL: E2E missing HTTPS connectivity test (OLS-3577 AC 2)
E2E verifies the MCP Deployment and Service objects exist with correct metadata, but never tests that app-server can actually reach the MCP server over HTTPS. A `kubectl exec` into the app-server pod curling `https://openshift-mcp-server:8443/healthz` with the mounted CA would cover this.

#### FAIL: E2E missing functional MCP tool test (OLS-3577 AC 3)
No test exercises MCP tool calls through the standalone service end-to-end.

#### FAIL: E2E missing cleanup verification (OLS-3577 AC 4)
No test toggles `introspectionEnabled` to false and verifies MCP resources (Deployment, Service, ConfigMaps, SA, NetworkPolicy) are deleted.

---

### Code Quality Issues

#### 1. `UpdateDeployment` + `Restart()` flow is fragile (should-fix)
**`internal/controller/ocpmcp/deployment.go:185-212`**

`UpdateDeployment` replaces `existingDeployment.Spec = desiredDeployment.Spec`, then calls `Restart()` which mutates `dep.Spec.Template.Annotations` on the *same in-memory object* and calls `r.Update()`. This works because Go map assignment is by reference, but the interaction is non-obvious. A clearer approach: either set the `forceReload` annotation directly in `UpdateDeployment` before the single `Update()` call, or have `Restart()` always fetch fresh from the API server.

#### 2. NetworkPolicy too permissive (should-fix)
**`internal/controller/ocpmcp/assets.go:148-153`**

The NetworkPolicy uses an empty `PodSelector` — any pod in the operator namespace can connect to the MCP server on port 8443. Only the app-server should be allowed. Restrict to:
```go
PodSelector: &metav1.LabelSelector{
    MatchLabels: map[string]string{
        "app.kubernetes.io/name": "lightspeed-app-server",
    },
},
```

#### 3. Narrow CA hash bootstrapping window (should-fix)
**`internal/controller/utils/mcp_server_config.go:59-74`**

When introspection is first enabled, the CA ConfigMap may exist but service-ca hasn't injected `service-ca.crt` yet. `GetOpenShiftMCPServerCACertHash` returns `""`, which is written to the app-server annotation. The empty→populated transition IS detected on the next reconcile, so this self-heals — but there's a window where the app-server has the CA volume mounted with no data, causing MCP TLS errors. Consider requeueing after a short delay when the hash is empty and introspection is enabled, rather than writing the empty hash.

#### 4. `guardrails = "none"` undocumented (nice-to-have)
**`internal/controller/ocpmcp/assets.go:40`**

The metrics toolset config disables guardrails. The security rationale should be documented — is this because the user's SA token already limits what they can query?

blublinsky:
@vimalk78 Re: **E2E missing HTTPS connectivity test ([OLS-3577](https://redhat.atlassian.net/browse/OLS-3577) AC 2)**

App-server is already wired to the standalone MCP Service over HTTPS (`https://openshift-mcp-server.<ns>.svc:8443/mcp`) with the service-ca CA mount. Existing e2e paths that exercise OpenShift MCP / introspection therefore already run against that HTTPS endpoint — a separate `kubectl exec` + `curl /healthz` would only re-check TLS, not the real client path used in production.

Happy to add an explicit healthz probe if you still want a belt-and-suspenders assertion, but I don't think AC 2 is uncovered given the HTTPS wiring + existing MCP e2e coverage.

blublinsky:
@vimalk78 Re: **E2E missing functional MCP tool test ([OLS-3577](https://redhat.atlassian.net/browse/OLS-3577) AC 3)**

Same rationale as AC 2: existing e2e that exercises OpenShift MCP / introspection tools already goes through the standalone HTTPS Service (app-server → `https://openshift-mcp-server.<ns>.svc:8443/mcp`). Those paths cover tool function end-to-end under the new wiring; a separate dedicated tool-call case wouldn’t add coverage beyond what’s already exercised.

Happy to strengthen assertions later if we want an explicit “tool was used” check, but I don’t see AC 3 as uncovered for this PR.

blublinsky:
@vimalk78 Re: **E2E missing cleanup verification ([OLS-3577](https://redhat.atlassian.net/browse/OLS-3577) AC 4)**

We’re not adding new e2e cases in this PR. Cleanup when `introspectionEnabled` is false is already covered by unit tests in `internal/controller/ocpmcp` (resources removed on disable). Happy to track a follow-up e2e for the toggle if we want cluster-level coverage later.

blublinsky:
@vimalk78 Re: **`UpdateDeployment` + `Restart()` flow**

Agreed — fixed. `UpdateDeployment` now stamps `force-reload` and persists with a single `Update()`. `Restart()` (TLS watcher path) always re-fetches the Deployment from the API instead of mutating a shared in-memory object.

blublinsky:
@vimalk78 Re: **NetworkPolicy too permissive**

Intentional. Ingress from any pod in the operator namespace on `:8443` is by design so future in-namespace clients (notably agentic sandbox) can reach the MCP Service without NP churn. App-server is the current consumer; sandboxes will follow the same trust path. Documented in `.ai/spec/what/ocpmcp.md`. We can tighten later if we decide sandboxes should go through a different network path.

blublinsky:
@vimalk78 Re: **Narrow CA hash bootstrapping window**

Agreed — fixed. When introspection is enabled and `openshift-mcp-server-ca` is missing or `service-ca.crt` has not been injected yet, `GetOpenShiftMCPServerCACertHash` now returns `ErrOpenShiftMCPServerCANotReady` instead of `""`. App-server generate/update fail and reconcile requeues, so we don’t stamp an empty hash annotation during the service-ca inject window. Once the cert appears, the next reconcile proceeds with a real hash and rolls as needed.

blublinsky:
@vimalk78 Re: **`guardrails = "none"` undocumented**

Clarified and tightened. `guardrails` is PromQL query-safety (cardinality / blanket regex / etc.), not Kubernetes RBAC — auth still uses the caller’s bearer token forwarded to Thanos/Alertmanager.

We switched from `"none"` to `"!tsdb"`, matching upstream OpenShift guidance when Thanos Querier lacks `/api/v1/status/tsdb`, while leaving the other query-safety checks enabled. Documented in the TOML comment and `.ai/spec/what/ocpmcp.md`.

vimalk78 (commented):
## Round 2 Adversarial Review — Score: 72/100

**Mode:** Jira (OLS-3574, OLS-3575, OLS-3576, OLS-3577) | **Round:** 2

### Adherence: 16/20 AC passed, 0 failed, 4 need human review

Score improved from 62 → 72. Two code quality issues from round 1 are resolved; two remain.

### Resolved from Round 1

- **UpdateDeployment + Restart() flow** — Fixed. `UpdateDeployment` mutates the in-memory deployment (spec + annotations) and passes it to `Restart()`, which stamps `force-reload` and issues a single `r.Update()`. No double-update risk. `Restart()` re-fetches from the API when called standalone (TLS watcher path). Clean fix.

- **NetworkPolicy breadth** — Accepted as intentional. Documented in code comment (`assets.go:131-132`) and spec (`ocpmcp.md`). Future sandbox consumers justify the open `PodSelector`.

### Still Open from Round 1

1. **`guardrails = "none"` not changed to `"!tsdb"`** (`internal/controller/ocpmcp/assets.go:40`)
   PR comments state this was changed to `"!tsdb"` and documented, but the code still shows `guardrails = "none"` with no documentation comment. Either the fix wasn't pushed or was reverted.

2. **CA hash bootstrapping fix not landed** (`internal/controller/utils/mcp_server_config.go:52-73`)
   PR comments state `GetOpenShiftMCPServerCACertHash` now returns `ErrOpenShiftMCPServerCANotReady` when CA is missing. The function still returns `("", nil)` for all CA-not-available scenarios — no sentinel error exists in the codebase. The existing behavior (empty→populated annotation comparison triggers restart on next reconcile) mitigates this to a narrow timing window, so downgraded from should-fix to nice-to-have.

### E2E Coverage (ACs 18-20) — Needs Human Review

Downgraded from FAIL to NEEDS_REVIEW. Author argues existing e2e queries implicitly exercise the standalone MCP HTTPS path. Analysis shows:
- `introspectionEnabled=true` in e2e CR (`assets.go:238`) ✓
- Standalone MCP server is deployed and verified (Test 7) ✓
- App-server config points to `https://openshift-mcp-server.<ns>.svc:8443/mcp` ✓
- **However**, basic queries like "what is OpenShift?" are answered by the LLM without tool calls — MCP tools are only invoked when the LLM decides to use cluster introspection, which is non-deterministic

Whether implicit coverage is sufficient is a judgment call for the human reviewer. Dedicated MCP connectivity / tool-call / cleanup toggle tests would make this deterministic.

---

Round 2 is the final AI review. Remaining items are for the human reviewer.

blublinsky:
@vimalk78 Re: Round 2 — items marked still open

Both fixes are on the current PR head (`e34891dd`):

1. **`guardrails`** — `internal/controller/ocpmcp/assets.go` now uses `guardrails = "!tsdb"` with an inline comment explaining PromQL query-safety (not RBAC) and why TSDB-dependent checks are disabled for OpenShift Thanos. Also noted in `.ai/spec/what/ocpmcp.md`.

2. **CA hash bootstrapping** — `GetOpenShiftMCPServerCACertHash` returns `ErrOpenShiftMCPServerCANotReady` when the CA ConfigMap is missing or `service-ca.crt` is empty; app-server generate/update fail and reconcile requeues instead of stamping an empty hash.

If the Files tab still showed the old versions, please refresh against that commit — happy to point at the exact hunks if useful.

On E2E ACs 18–20: agreed this is a judgment call. Existing e2e already deploys standalone MCP over HTTPS and runs with `introspectionEnabled=true`; dedicated connectivity / tool-call / disable-cleanup cases remain optional follow-ups rather than blockers for this cutover PR.

blublinsky:
@vimalk78 Follow-up with exact locations on `e34891dd`:

**1. `guardrails = "!tsdb"`**
- [`internal/controller/ocpmcp/assets.go`](https://github.com/blublinsky/lightspeed-operator/blob/e34891dd329957794c9c8b284bf767a63c16b511/internal/controller/ocpmcp/assets.go#L40-L43) — comment + `guardrails = "!tsdb"` (L40–L43)
- [`.ai/spec/what/ocpmcp.md`](https://github.com/blublinsky/lightspeed-operator/blob/e34891dd329957794c9c8b284bf767a63c16b511/.ai/spec/what/ocpmcp.md#L51) — Security rule 17 (L51)

**2. CA hash bootstrapping**
- [`internal/controller/utils/errors.go`](https://github.com/blublinsky/lightspeed-operator/blob/e34891dd329957794c9c8b284bf767a63c16b511/internal/controller/utils/errors.go#L207) — `ErrOpenShiftMCPServerCANotReady` (L207)
- [`internal/controller/utils/mcp_server_config.go`](https://github.com/blublinsky/lightspeed-operator/blob/e34891dd329957794c9c8b284bf767a63c16b511/internal/controller/utils/mcp_server_config.go#L50-L71) — `GetOpenShiftMCPServerCACertHash` returns that error when the CA ConfigMap is missing or `service-ca.crt` is empty (L50–L71)
- [`internal/controller/appserver/deployment.go`](https://github.com/blublinsky/lightspeed-operator/blob/e34891dd329957794c9c8b284bf767a63c16b511/internal/controller/appserver/deployment.go#L494-L498) — generate path fails/requeues on error (L494–L498)
- [`internal/controller/appserver/deployment.go`](https://github.com/blublinsky/lightspeed-operator/blob/e34891dd329957794c9c8b284bf767a63c16b511/internal/controller/appserver/deployment.go#L675-L681) — update path fails/requeues on error (L675–L681)

PR #1858, opened 2026-07-22
OLS-3705 Fix OTel collector crash: routing/traces table must not be empty

vimalk78:
## Summary

- The routing connector requires at least one table entry — `routing/traces` was configured with only `default_pipelines` and an empty table, causing the collector to crash on startup with: `invalid routing table: the routing table is empty`
- Add a routing rule for `lightspeed-service` traces, matching the existing `routing/logs` pattern
- Add `OtelAppServerServiceName` constant for the trace routing condition

## Test plan

- [x] `make test` — all unit tests pass
- [x] New assertion verifies `routing/traces` table is non-empty
- [ ] Deploy and confirm collector starts without crash

## Summary by CodeRabbit

* **Bug Fixes**
  * When a tracing endpoint is configured, the collector now uses the in-cluster service CA bundle for secure OTLP TLS connections.
  * Improved trace routing so only traces from the Lightspeed service are sent to the Lightspeed trace pipeline, with routing no longer enabled by default.

* **Tests**
  * Expanded coverage to validate the generated trace-routing connector configuration by parsing the ConfigMap YAML and asserting the connector’s routing table structure.

### DIFF
--- a/internal/controller/otelcollector/assets.go
+++ b/internal/controller/otelcollector/assets.go
@@ -379,7 +379,13 @@ func buildCollectorConfigYAML(cr *olsv1alpha1.OLSConfig) ([]byte, error) {
 			},
 		}
 		connectors["routing/traces"] = map[string]interface{}{
-			"default_pipelines": []interface{}{"traces/lightspeed"},
+			"default_pipelines": []interface{}{},
+			"table": []interface{}{
+				map[string]interface{}{
+					"condition": fmt.Sprintf(`resource.attributes["service.name"] == %q`, utils.OtelAppServerServiceName),
+					"pipelines": []interface{}{"traces/lightspeed"},
+				},
+			},
 		}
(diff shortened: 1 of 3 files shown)


blublinsky (commented):
Soft recommendations (not blockers — fix LGTM for OLS-3705):

1. **Test strength** — the new assert only checks that `routing/traces` `table` is non-empty. Consider also asserting the condition, target pipeline (`traces/lightspeed`), empty `default_pipelines`, and/or use of `OtelAppServerServiceName` (closer to how `routing/logs` is covered).

2. **Forwarding semantics** — emptying `default_pipelines` and filtering on `service.name == lightspeed-service` drops unmatched traces (vs the previous intent of defaulting everything to `traces/lightspeed`). That looks intentional and matches the logs pattern + what lightspeed-service emits; just worth an explicit ACK that dropping non-matching traces is desired.

Otherwise looks good — thanks for the fix.

PR #1870, opened 2026-07-24
OLS-3752 - Removing agentic operator image references.

sriroopar:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update

- Related Issue #
- Closes #

## Testing

## Summary by CodeRabbit

* **Configuration**
  * Updated operator image arguments and bundle related image metadata to shift from agentic console/alerts-adapter images to OTEL Collector and RHOKP.
  * Adjusted default image resolution to pull directly from bundle-related defaults.
* **New Features**
  * Agentic console UI and alerts adapter reconciliation is now gated by whether their image values are provided (and alerts adapter config reference is enabled).
* **Bug Fixes**
  * Readiness/status now consistently reports disabled components with clear reasons when images/config are missing.
* **Tests**
  * Updated expectations for generated deployment images and added coverage for disabled-when-image-empty gating behavior.

### DIFF
--- a/internal/controller/agenticconsole/suite_test.go
+++ b/internal/controller/agenticconsole/suite_test.go
@@ -105,12 +105,14 @@ var _ = BeforeSuite(func() {
 		Expect(k8sClient.Create(ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: ns}})).To(Succeed())
 	}
 
-	testReconcilerInstance = utils.NewTestReconciler(
+	tr := utils.NewTestReconciler(
 		k8sClient,
 		logf.Log.WithName("controller").WithName("OLSConfig"),
 		scheme.Scheme,
 		utils.OLSNamespaceDefault,
 	)
+	tr.AgenticConsoleImage = "registry.redhat.io/openshift-lightspeed/lightspeed-agentic-console-rhel9:latest"
+	testReconcilerInstance = tr
 
 	cr = &olsv1alpha1.OLSConfig{}
 	crNamespacedName = types.NamespacedName{Name: "cluster"}
--- a/internal/controller/olsconfig_controller.go
+++ b/internal/controller/olsconfig_controller.go
@@ -302,18 +302,13 @@ func (r *OLSConfigReconciler) reconcileOperatorResources(ctx context.Context) er
 // Agentic integration handoff runs at the end of Phase 2 (after Services/TLS).
 // Uses continue-on-error to reconcile as many resources as possible, even if some fail.
 func (r *OLSConfigReconciler) reconcileIndependentResources(ctx context.Context, olsconfig *olsv1alpha1.OLSConfig) error {
+	resourceFailures := make(map[string]error)
 	resourceSteps := []utils.ReconcileSteps{
-		{Name: "postgres resources", Fn: func(ctx context.Context, cr *olsv1alpha1.OLSConfig) error {
-			return postgres.ReconcilePostgresResources(r, ctx, cr)
-		}},
 		{Name: "console UI resources", Fn: func(ctx context.Context, cr *olsv1alpha1.OLSConfig) error {
 			return console.ReconcileConsoleUIResources(r, ctx, cr)
 		}},
-		{Name: "agentic console UI resources", Fn: func(ctx context.Context, cr *olsv1alpha1.OLSConfig) error {
-			return agenticconsole.ReconcileAgenticConsoleUIResources(r, ctx, cr)
-		}},
-		{Name: "alerts adapter resources", Fn: func(ctx context.Context, cr *olsv1alpha1.OLSConfig) error {
-			return alertsadapter.ReconcileAlertsAdapterResources(r, ctx, cr)
+		{Name: "postgres resources", Fn: func(ctx context.Context, cr *olsv1alpha1.OLSConfig) error {
+			return postgres.ReconcilePostgresResources(r, ctx, cr)
 		}},
 		{Name: "OTEL Collector resources", Fn: func(ctx context.Context, cr *olsv1alpha1.OLSConfig) error {
 			return otelcollector.ReconcileOtelCollectorResources(r, ctx, cr)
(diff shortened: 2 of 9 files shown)


internal/controller/olsconfig_controller.go
```diff
@@ -318,16 +315,37 @@ func (r *OLSConfigReconciler) reconcileIndependentResources(ctx context.Context,
...
 	}
 
+	if r.Options.AgenticConsoleUIImage != "" {
+		resourceSteps = append(resourceSteps, utils.ReconcileSteps{
+			Name: "agentic console UI resources",
+			Fn: func(ctx context.Context, cr *olsv1alpha1.OLSConfig) error {
+				return agenticconsole.ReconcileAgenticConsoleUIResources(r, ctx, cr)
+			},
+		})
+	} else {
```
xrajesh: This cleanup block runs on **every reconcile loop** when the image is empty, not just on the first transition. Since neither resource exists in a fresh install, this fires API calls to the cluster that always return NotFound. `RemoveAgenticConsole` is idempotent so it won't cause errors, but it's noisy and wasteful.

Same concern applies to `RemoveAlertsAdapter` a few lines below (position 34).

Simplest fix: wrap each in a check against the existing status condition to detect a prior enabled state before calling remove, so this only fires on actual transition.

internal/controller/olsconfig_controller.go
```diff
@@ -442,22 +457,56 @@ func (r *OLSConfigReconciler) reconcileDeploymentsAndStatus(ctx context.Context,
...
 			Fn: func(ctx context.Context, cr *olsv1alpha1.OLSConfig) error {
-				return alertsadapter.ReconcileAlertsAdapterDeployment(r, ctx, cr)
+				return agenticconsole.ReconcileAgenticConsoleUIDeploymentAndPlugin(r, ctx, cr)
 			},
-			ConditionType: utils.TypeAlertsAdapterReady,
-			Deployment:    utils.AlertsAdapterDeploymentName,
+			ConditionType: utils.TypeAgenticConsolePluginReady,
+			Deployment:    utils.AgenticConsoleUIDeploymentName,
 		})
+	} else {
```
xrajesh, line 472: `ConditionTrue` with reason `NotConfigured` is semantically misleading — a status watcher or the console UI will read `AgenticConsolePluginReady=True` as "component is healthy and running". The component is intentionally absent.

Prefer `ConditionFalse` + reason `Disabled`, or skip setting the condition entirely for components that are not deployed. Same applies to the two `TypeAlertsAdapterReady` conditions set further below with the same pattern.

internal/controller/agenticconsole/suite_test.go
```diff
@@ -105,12 +105,14 @@ var _ = BeforeSuite(func() {
...
 	}
 
-	testReconcilerInstance = utils.NewTestReconciler(
+	tr := utils.NewTestReconciler(
 		k8sClient,
 		logf.Log.WithName("controller").WithName("OLSConfig"),
 		scheme.Scheme,
 		utils.OLSNamespaceDefault,
 	)
+	tr.AgenticConsoleImage = "registry.redhat.io/openshift-lightspeed/lightspeed-agentic-console-rhel9:latest"
```
xrajesh, line 114: This updates the asset-generation test to use an explicit image, which is correct. But there's no test covering the **gate behavior** itself: when `AgenticConsoleUIImage == ""`, the reconcile steps should be skipped and no `Deployment`, `ConsolePlugin`, or `ClusterRole` should be created.

That's the central correctness claim of this PR and it's currently untested. A controller integration test with an empty image config would catch any future regression.

xrajesh (commented):
Core logic is correct — gating on image presence is the right approach and the CSV/related_images cleanup is complete. Three things worth addressing before merge.

xrajesh:
**Process:** PR description is empty and there's no Jira ticket in the title. Please fill in the checklist and add `OLS-XXXX` to the title so this is traceable.

internal/controller/olsconfig_reconciler_test.go
```diff
@@ -258,4 +261,138 @@ var _ = Describe("OLSConfig Reconciler Helper Functions", Ordered, func() {
...
+				_ = emptyImageReconciler.reconcileIndependentResources(ctx, cr)
+
+				dep := &appsv1.Deployment{}
+				err := k8sClient.Get(ctx, types.NamespacedName{
+					Name:      utils.AgenticConsoleUIDeploymentName,
+					Namespace: namespace,
+				}, dep)
+				Expect(apierrors.IsNotFound(err)).To(BeTrue(),
+					"agentic console Deployment should not exist when image is empty")
+			})
```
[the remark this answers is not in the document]
raptorsun: code rabbit is right, the test of deployments should be tested  after emptyImageReconciler.reconcileDeploymentsAndStatus(ctx, cr)

internal/controller/olsconfig_reconciler_test.go
```diff
@@ -258,4 +261,138 @@ var _ = Describe("OLSConfig Reconciler Helper Functions", Ordered, func() {
...
+					Namespace: namespace,
+				}, sa)
+				Expect(apierrors.IsNotFound(err)).To(BeTrue(),
+					"agentic console ServiceAccount should not exist when image is empty")
+			})
+		})
+
+		Context("when AlertsAdapterImage is empty", func() {
+			It("should not create alerts adapter Deployment", func() {
+				_ = emptyImageReconciler.reconcileIndependentResources(ctx, cr)
```
raptorsun: the test of deployment should be done after reconcileIndependentResources.

raptorsun: the independent resources are step 1.
the deployments are step 2 in the reconciliation loop.

raptorsun:
we need to fix the test on deployments creation as code rabbit found.
another review on "Use configMapRef to gate alerts-adapter resources in both phases." worth our attention, too. the reconcialiation is 2 phases, 1st to create the surrounding resources and the 2nd is to start the deployments. we need verify in both steps the agentic component is disable when missing the image.

raptorsun:
/override "Red Hat Konflux / bundle-e2e-tests / ols-bundle"
/override "Red Hat Konflux / service-e2e-tests-419 / ols-bundle"
we need to update the image list in the bundle soon once this PR is merged and the operator component released

PR #1897, opened 2026-07-30
OLS-3593 enable init container comparison for app server deployment

raptorsun:
## Summary
- Fix: detect changes to `.spec.ols.rag` (RAG image references) so they trigger a deployment rollout
- Root cause: `DeploymentSpecEqual` was called with `compareInitContainers=false`, and no other change detector tracked RAG images — so RAG image changes were silently ignored

## Why not just enable init container comparison?

The obvious fix — flipping `compareInitContainers` from `false` to `true` — causes an **infinite reconciliation loop** when RAG is configured.

When RAG images are present, the operator sets an `image.openshift.io/triggers` annotation on the deployment (`rag.go:62-79`). This tells OpenShift's image trigger controller to resolve ImageStreamTags and replace the init container image field with the resolved `@sha256:...` digest. With `compareInitContainers=true`, this creates a fight between two controllers:

1. **Operator reconciles** → generates desired deployment with the user-specified image tag (e.g., `my-image:latest`)
2. **OpenShift trigger controller** → resolves the tag and writes the SHA digest into the deployment (e.g., `my-image@sha256:abc...`)
3. **Operator reconciles again** → `ContainerSpecEqual` compares `a.Image == b.Image`, sees tag != digest, detects a "change"
4. **Operator updates** the deployment back to the tag → goto step 2

This was confirmed in e2e testing where the deployment generation reached 5663 vs observed generation 2303 (~5-6 updates/second).

## Solution

Track a SHA-256 hash of the CR's `.spec.ols.rag` in a deployment annotation (`ols.openshift.io/rag-spec-hash`). When the user changes RAG image references, the hash changes and triggers a deployment rollout — without comparing init container images directly and without conflicting with OpenShift's image trigger mechanism.

## Test plan
- [x] `make test` passes — all unit tests green
- [x] `bundle-e2e-4-21` CI passed (including BYOK/RAG tests)
- [ ] Deploy OLS with a RAG image configured, change the image tag, and verify the operator rolls out a new deployment

Fixes: [OLS-3593](https://redhat.atlassian.net/browse/OLS-3593)

### DIFF
--- a/internal/controller/appserver/deployment.go
+++ b/internal/controller/appserver/deployment.go
@@ -2,6 +2,7 @@ package appserver
 
 import (
 	"context"
+	"crypto/sha256"
 	"encoding/json"
 	"fmt"
 	"path"
@@ -466,6 +467,7 @@ func GenerateOLSDeployment(r reconciler.Reconciler, cr *olsv1alpha1.OLSConfig) (
 
 	annotations := map[string]string{
 		utils.OLSConfigMapResourceVersionAnnotation: configMapResourceVersion,
+		utils.RAGSpecHashAnnotation:                 ragSpecHash(cr),
 		utils.ProxyCACertHashAnnotation:             proxyCACMResourceVersion,
 	}
 
@@ -616,6 +618,13 @@ func updateOLSDeployment(r reconciler.Reconciler, ctx context.Context, cr *olsv1
 		}
 	}
 
+	// Step 4: Check if RAG spec has changed
+	currentRAGHash := ragSpecHash(cr)
+	if existingDeployment.Annotations[utils.RAGSpecHashAnnotation] != currentRAGHash {
+		r.GetLogger().Info("RAG spec changed, updating deployment")
+		changed = true
+	}
+
 	// If nothing changed, skip update
 	if !changed {
 		return nil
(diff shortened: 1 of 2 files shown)


blublinsky (commented):
## Review: OLS-3593 RAG spec hash detection

Nice, clean fix for a subtle bug. The hash-based approach avoids the image trigger controller fight elegantly.

### Non-blocking observations

1. **`json.Marshal` error discarded** (`deployment.go:ragSpecHash()`) — `data, _ := json.Marshal(...)` drops the error. In practice `RAGSpec` is a simple struct and won't fail, but a `//nolint:errcheck` comment or a brief rationale would help future readers. Not blocking since `sha256.Sum256(nil)` is deterministic (no spurious rollouts).

2. **No unit test for the new logic** — a table-driven test for `ragSpecHash()` (empty / single / multiple entries) and one case in `updateOLSDeployment` verifying the annotation-based rollout would be straightforward additions. Not blocking since e2e coverage exists.

Both are nice-to-haves. LGTM otherwise — the approach is sound and the PR description is excellent.

raptorsun:
/test bundle-e2e-4-21
CI could not claim a cluster

PR #1921, opened 2026-08-05
OLS-3799 Add wait-for-rhokp init container to app-server deployment

vimalk78:
## Summary
- Adds a `wait-for-rhokp` init container to the app-server Deployment when `ByokRAGOnly=false`
- Polls the RHOKP/Solr HTTPS endpoint with curl using exponential backoff (360s timeout, matching RHOKP's startup probe budget)
- Prevents the app-server from starting before Solr is reachable, fixing a race condition where `SolrHybridSearch`'s `@cached_property` permanently caches `None` if RHOKP takes longer than the 120s retry window
- Follows the existing `wait-for-postgres` init container pattern

## Test plan
- [x] Unit tests for `GenerateRHOKPWaitInitContainer` (image, security context, CA mount, curl command, timeout, namespace)
- [x] Deployment tests: init container present when `ByokRAGOnly=false`, absent when `ByokRAGOnly=true`
- [x] Existing `assets_test.go` `ConsistOf` assertion updated for the new init container
- [x] `make test` passes (130/130 specs)
- [ ] E2e: deploy with `ByokRAGOnly=false`, verify init container blocks app-server startup until RHOKP is serving

## Summary by CodeRabbit

* **New Features**
  * Added RHOKP/Solr readiness checks during RAG deployment startup.
  * Securely verifies RHOKP availability using mounted certificates and bounded retry behavior.
  * BYOK-only deployments skip the RHOKP readiness step.

* **Bug Fixes**
  * Prevents RAG components from starting before RHOKP is available.
  * Improves deployment reliability when RHOKP requires additional time to become ready.

### DIFF
--- a/internal/controller/utils/rhokp_wait.go
+++ b/internal/controller/utils/rhokp_wait.go
@@ -1,0 +1,92 @@
+package utils
+
+import (
+	"fmt"
+	"path"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+// RHOKPWaitMaxSeconds is the maximum time the wait-for-rhokp init container
+// will poll before giving up.  Matches the RHOKP startup-probe budget
+// (20 s initial-delay + 34 failures × 10 s period ≈ 360 s).
+const RHOKPWaitMaxSeconds = 360
+
+// GenerateRHOKPWaitInitContainer returns an init container that blocks until the
+// standalone RHOKP/Solr instance is reachable on its HTTPS endpoint.
+//
+// Without this gate the app-server's SolrHybridSearch initializer may exhaust
+// its 120 s retry budget before RHOKP is ready, permanently caching a nil
+// reference for product-docs RAG (OLS-3799).
+//
+// The container uses the app-server image (which ships curl) and polls the
+// Solr admin/ping endpoint.  The RHOKP CA certificate volume must already
+// be defined on the pod (AppRHOKPCACertVolumeName).
+func GenerateRHOKPWaitInitContainer(image, namespace string) corev1.Container {
+	rhokpURL := RHOKPServiceURL(namespace) + RHOOKPReadinessHTTPPath
+
+	caPath := path.Join(OLSAppCertsMountRoot, AppRHOKPCACertDir, AppRHOKPCACertFile)
+
+	script := fmt.Sprintf(`
+if ! command -v curl >/dev/null 2>&1; then
+  echo "wait-for-rhokp: curl not found in image" >&2
+  exit 1
+fi
+
+sleep_sec=1
+max_sleep=30
+start=$(date +%%s)
+max_elapsed=%d
+url="%s"
+ca="%s"
+
+backoff() {
+  sleep "$sleep_sec"
+  sleep_sec=$((sleep_sec * 2))
+  [ "$sleep_sec" -gt "$max_sleep" ] && sleep_sec="$max_sleep"
+}
+
+while true; do
+  now=$(date +%%s)
+  elapsed=$((now - start))
+  if [ "$elapsed" -ge "$max_elapsed" ]; then
+    echo "wait-for-rhokp: timed out after ${max_elapsed}s" >&2
+    exit 1
+  fi
+
+  if curl -sf --cacert "$ca" --max-time 5 "$url" >/dev/null 2>&1; then
+    echo "wait-for-rhokp: RHOKP/Solr is reachable"
+    exit 0
+  fi
+
+  echo "wait-for-rhokp: not ready yet (elapsed=${elapsed}s)" >&2
+  backoff
+done
+`, RHOKPWaitMaxSeconds, rhokpURL, caPath)
+
+	return corev1.Container{
+		Name:            RHOKPWaitInitContainerName,
+		Image:           image,
+		ImagePullPolicy: corev1.PullIfNotPresent,
+		Command:         []string{"/bin/sh", "-c", script},
+		VolumeMounts: []corev1.VolumeMount{
+			{
+				Name:      AppRHOKPCACertVolumeName,
+				MountPath: path.Join(OLSAppCertsMountRoot, AppRHOKPCACertDir),
+				ReadOnly:  true,
+			},
+		},
+		SecurityContext: RestrictedContainerSecurityContext(),
+		Resources: corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("10m"),
+				corev1.ResourceMemory: resource.MustParse("32Mi"),
+			},
+			Limits: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("100m"),
+				corev1.ResourceMemory: resource.MustParse("64Mi"),
+			},
+		},
+	}
+}
--- a/internal/controller/utils/rhokp_wait_test.go
+++ b/internal/controller/utils/rhokp_wait_test.go
@@ -1,0 +1,79 @@
+package utils
+
+import (
+	"fmt"
+	"path"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+var _ = Describe("RHOKP wait", func() {
+
+	Describe("GenerateRHOKPWaitInitContainer", func() {
+		const testNS = "openshift-lightspeed"
+
+		It("uses the provided image and generates correct container spec", func() {
+			c := GenerateRHOKPWaitInitContainer(OLSAppServerImageDefault, testNS)
+			Expect(c.Name).To(Equal(RHOKPWaitInitContainerName))
+			Expect(c.Image).To(Equal(OLSAppServerImageDefault))
+			Expect(c.SecurityContext).To(Equal(RestrictedContainerSecurityContext()))
+			Expect(c.Resources.Requests).To(HaveKey(corev1.ResourceCPU))
+			Expect(c.Resources.Requests).To(HaveKey(corev1.ResourceMemory))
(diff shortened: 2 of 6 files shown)


blublinsky (commented):
**Must-fix: `cr.Namespace` is always empty for the cluster-scoped `OLSConfig` CRD**

At `deployment.go:447`:
```go
initContainers = append(initContainers, utils.GenerateRHOKPWaitInitContainer(r.GetAppServerImage(), cr.Namespace))
```

`OLSConfig` is `scope: Cluster`, so `cr.Namespace` is always `""`. This produces a URL like `https://lightspeed-rhokp..svc:8443/...` (double dots — invalid DNS).

**Suggested fix:** Drop the `namespace` parameter entirely and use the short service name, matching the Postgres wait pattern. Since the init container runs in the same namespace as the RHOKP service, Kubernetes DNS resolves the short name automatically:

```go
rhokpURL := fmt.Sprintf("https://%s:%d/solr/portal-rag/select?q=*:*&rows=0&wt=json",
    RHOKPServiceName, RHOOKPImageHTTPSPort)
```

This also simplifies `GenerateRHOKPWaitInitContainer` — it no longer needs the namespace argument, just like `GeneratePostgresWaitInitContainer` only takes the image.

Note: `--cacert` TLS verification with a short hostname may require the service-ca cert to include the short name as a SAN. If FQDN is needed for TLS cert matching, use the constant `utils.OLSNamespaceDefault` instead of `cr.Namespace`.

blublinsky (commented):
**Questioning the approach: should this fix live in the operator at all?**

The app-server (lightspeed-service-api) already has retry logic for Solr connectivity. The root cause described here is that `SolrHybridSearch` uses `@cached_property` which permanently caches `None` on failure — that's a bug in the service code, not a deployment ordering problem.

The proper fix is in lightspeed-service-api: don't permanently cache a transient failure. For example, use a lazy property that retries on `None`, or replace `@cached_property` with a retry-aware initializer. This makes the service resilient regardless of startup ordering, container restarts, or network blips — situations an init container cannot help with anyway.

Adding an init container with shell scripts, CA mounts, and backoff logic in the operator is treating the symptom at the wrong layer. Compare with the Postgres wait: that exists because the app literally cannot start without a DB connection. Solr/RHOKP is not a hard prerequisite — the app should degrade gracefully and recover when it becomes available.

I'd suggest fixing the caching bug in the service and closing this PR. Happy to discuss if there's additional context I'm missing.

internal/controller/utils/rhokp_wait.go
```diff
@@ -0,0 +1,91 @@
...
+url="%s"
+ca="%s"
+
+backoff() {
+  sleep "$sleep_sec"
+  sleep_sec=$((sleep_sec * 2))
+  [ "$sleep_sec" -gt "$max_sleep" ] && sleep_sec="$max_sleep"
+}
+
+while true; do
```
vimalk78, line 50: **Missing `command -v curl` guard**

`postgres_wait.go` validates its tool exists before entering the poll loop:
```sh
if ! command -v pg_isready >/dev/null 2>&1; then
  echo "wait-for-postgres: pg_isready not found in image" >&2
  exit 1
fi
```

This script has no equivalent check for `curl`. If curl is ever removed from the app-server image, the init container will loop for 360 seconds (the curl invocation suppresses all output via `>/dev/null 2>&1`) before timing out — a confusing silent failure instead of a fast one.

Suggestion: add `command -v curl` check before the `while true` loop, matching the postgres pattern.

internal/controller/utils/rhokp_wait.go
```diff
@@ -0,0 +1,91 @@
...
+		VolumeMounts: []corev1.VolumeMount{
+			{
+				Name:      AppRHOKPCACertVolumeName,
+				MountPath: path.Join(OLSAppCertsMountRoot, AppRHOKPCACertDir),
+				ReadOnly:  true,
+			},
+		},
+		SecurityContext: &corev1.SecurityContext{
+			AllowPrivilegeEscalation: &[]bool{false}[0],
+			ReadOnlyRootFilesystem:   &[]bool{true}[0],
```
vimalk78: **Incomplete SecurityContext — missing PSA "restricted" fields**

This only sets `AllowPrivilegeEscalation` and `ReadOnlyRootFilesystem`. The codebase has `RestrictedContainerSecurityContext()` (`utils/utils.go:73`) that also sets:
- `RunAsNonRoot: true`
- `SeccompProfile: RuntimeDefault`
- `Capabilities: Drop ALL`

These are required by the Pod Security "restricted" profile. Without them, the pod could fail admission on namespaces with PSA enforcement.

Note: `wait-for-postgres` has the same gap (pre-existing), but this PR should not compound it.

Suggestion: use `RestrictedContainerSecurityContext()` instead of the inline SecurityContext.

internal/controller/utils/rhokp_wait.go
```diff
@@ -0,0 +1,91 @@
...
+//
+// Without this gate the app-server's SolrHybridSearch initializer may exhaust
+// its 120 s retry budget before RHOKP is ready, permanently caching a nil
+// reference for product-docs RAG (OLS-3799).
+//
+// The container uses the app-server image (which ships curl) and polls a
+// lightweight Solr select query.  The RHOKP CA certificate volume must already
+// be defined on the pod (AppRHOKPCACertVolumeName).
+func GenerateRHOKPWaitInitContainer(image, namespace string) corev1.Container {
+	rhokpURL := fmt.Sprintf("https://%s.%s.svc:%d/solr/portal-rag/select?q=*:*&rows=0&wt=json",
```
vimalk78: **Readiness endpoint mismatch + duplicated URL construction**

Two issues on this line:

1. **Endpoint**: This polls `/solr/portal-rag/select?q=*:*&rows=0&wt=json`, but RHOKP's own readiness/startup probes use `RHOOKPReadinessHTTPPath` = `/solr/portal-rag/admin/ping`. The `select` endpoint is heavier — it queries the Solr index (even with `rows=0`). During startup, `/admin/ping` becomes available first. The init container should use the same endpoint as the readiness probe.

2. **URL helper**: There's an existing `RHOKPServiceURL(namespace)` helper (`constants.go:714`) that returns `https://lightspeed-rhokp.<namespace>.svc:8443`. This could be reused:
```go
rhokpURL := RHOKPServiceURL(namespace) + RHOOKPReadinessHTTPPath
```

vimalk78 (commented):
## Adversarial Review

Overall the approach is sound — gating on `!ByokRAGOnly` is consistent, the timeout matches the RHOKP startup probe budget, and the exponential backoff is well-implemented. Four issues found in inline comments below.

internal/controller/utils/rhokp_wait.go
vimalk78: Fixed — now uses `RHOKPServiceURL(namespace) + RHOOKPReadinessHTTPPath` (`/solr/portal-rag/admin/ping`). Lighter endpoint, available earlier during startup, and no duplicated URL construction.

internal/controller/utils/rhokp_wait.go
vimalk78: Added `command -v curl` guard before the poll loop, matching the postgres pattern. Fails fast with a clear message instead of silently looping for 360s.

internal/controller/utils/rhokp_wait.go
vimalk78: Switched to `RestrictedContainerSecurityContext()` — picks up RunAsNonRoot, SeccompProfile RuntimeDefault, and Drop ALL capabilities.

internal/controller/utils/rhokp_wait_test.go
```diff
@@ -0,0 +1,75 @@
...
+			Expect(c.SecurityContext).NotTo(BeNil())
+			Expect(*c.SecurityContext.AllowPrivilegeEscalation).To(BeFalse())
+			Expect(*c.SecurityContext.ReadOnlyRootFilesystem).To(BeTrue())
+			Expect(c.Resources.Requests).To(HaveKey(corev1.ResourceCPU))
+			Expect(c.Resources.Requests).To(HaveKey(corev1.ResourceMemory))
+			Expect(c.Resources.Requests[corev1.ResourceCPU]).To(Equal(resource.MustParse("10m")))
+			Expect(c.Resources.Requests[corev1.ResourceMemory]).To(Equal(resource.MustParse("32Mi")))
+			Expect(c.Resources.Limits[corev1.ResourceCPU]).To(Equal(resource.MustParse("100m")))
+			Expect(c.Resources.Limits[corev1.ResourceMemory]).To(Equal(resource.MustParse("64Mi")))
+			Expect(c.Command).To(HaveLen(3))
```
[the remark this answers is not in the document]
vimalk78: This is a generated init container spec — the shell script runs inside a Kubernetes pod, not in the test process. Unit-testing shell behavior (mocking curl, date, sleep, asserting exit codes and deadline boundaries) would require a full shell execution harness, which is disproportionate for a readiness probe script. The test pattern here matches `postgres_wait_test.go` — verify the generated container spec is correct (image, security context, volume mounts, resources) and that the script contains the expected URL, timeout, backoff, and guard logic. Actual behavior is validated by the init container running on a real cluster.

vimalk78 (commented):
Fixed — switched from `cr.Namespace` to `r.GetNamespace()` which returns the operator watch namespace. Kept the FQDN approach (with namespace) rather than short service name because the service-ca TLS cert SANs use the FQDN — `curl --cacert` would fail cert validation against a short hostname.

vimalk78:
Re @blublinsky review — fixed. Switched from `cr.Namespace` (always empty for cluster-scoped `OLSConfig`) to `r.GetNamespace()` which returns the operator watch namespace. Kept the FQDN rather than short service name because service-ca TLS certs use the FQDN as SAN — `curl --cacert` would fail cert validation against a short hostname.

raptorsun:
the fix to lightspeed service as @blublinsky suggested can be addressed in another PR to improve resiliancy of that component.

raptorsun:
/hold
pending decision over 2 solutions: this one and [the service fix](https://github.com/openshift/lightspeed-service/pull/3045)

raptorsun:
/close
in favor of https://github.com/openshift/lightspeed-service/pull/3045

PR #1954, opened 2026-08-13
OLS-3886 Tighten ClusterRole RBAC to prevent privilege escalation

xrajesh:
## Summary

- **Secrets**: Moved unrestricted cluster-wide secrets access (`create/delete/deletecollection/get/list/patch/update/watch`) from ClusterRole to namespace-scoped Role (`openshift-lightspeed`). Only `pull-secret` read access (pinned by `resourceNames`) remains cluster-scoped for telemetry.
- **RBAC objects**: Restricted `clusterroles`, `clusterrolebindings`, and `rolebindings` write access to specific `resourceNames` the operator manages (`lightspeed-app-server-sar-role`, `lightspeed-agentic-alerts-adapter-*`), eliminating the ability to bind `cluster-admin` or create arbitrary roles.
- **NetworkPolicies**: Moved from ClusterRole to namespace-scoped Role — all network policies are created in `openshift-lightspeed` only.

## Motivation

A compromised operator ServiceAccount token could previously:
1. Read every secret cluster-wide (TLS certs, registry creds, SA tokens)
2. Create a ClusterRoleBinding to `cluster-admin` for full cluster takeover
3. Delete network policies in any namespace to bypass isolation

## Test plan

- [x] `make manifests` regenerates `config/rbac/role.yaml` correctly
- [x] `make test` — all controller tests pass
- [ ] Deploy to a dev cluster and verify OLSConfig reconciliation succeeds
- [ ] Verify the operator can still read `pull-secret` from `openshift-config`
- [ ] Verify the operator can still create its named ClusterRoles/Bindings
- [ ] Verify network policies are created in `openshift-lightspeed`

Fixes: https://redhat.atlassian.net/browse/OLS-3886

## Summary by CodeRabbit

* **Security**
  * Restricted access to Secrets and NetworkPolicies within the `openshift-lightspeed` namespace.
  * Limited permissions to specifically required resources instead of broad cluster-wide access.
  * Removed obsolete and overly permissive access rules.
  * Prevented invalid permission combinations that could unintentionally broaden access.

* **Tests**
  * Added automated validation for access permissions, resource restrictions, required resources, and namespace placement.

### DIFF
--- a/internal/rbac/role_yaml_test.go
+++ b/internal/rbac/role_yaml_test.go
@@ -1,0 +1,198 @@
+package rbac
+
+import (
+	"bytes"
+	"os"
+	"path/filepath"
+	"runtime"
+	"sort"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+func loadRoleYAML() (rbacv1.ClusterRole, rbacv1.Role) {
+	_, thisFile, _, _ := runtime.Caller(0)
+	roleFile := filepath.Join(filepath.Dir(thisFile), "..", "..", "config", "rbac", "role.yaml")
+
+	data, err := os.ReadFile(roleFile)
+	Expect(err).NotTo(HaveOccurred(), "reading role.yaml")
+
+	decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(data), 4096)
+
+	var cr rbacv1.ClusterRole
+	Expect(decoder.Decode(&cr)).NotTo(HaveOccurred(), "decoding ClusterRole")
+
+	var r rbacv1.Role
+	Expect(decoder.Decode(&r)).NotTo(HaveOccurred(), "decoding Role")
+
+	return cr, r
+}
+
+var _ = Describe("manager-role YAML", func() {
+	var (
+		clusterRole rbacv1.ClusterRole
+		role        rbacv1.Role
+	)
+
+	BeforeEach(func() {
+		clusterRole, role = loadRoleYAML()
+	})
+
+	Describe("ClusterRole", func() {
+		It("restricts secrets access to exactly pull-secret", func() {
+			found := false
+			for _, rule := range clusterRole.Rules {
+				for _, res := range rule.Resources {
+					if res == "secrets" {
+						found = true
+						Expect(rule.APIGroups).To(ConsistOf(""),
+							"ClusterRole secrets rule must be in the core API group")
+						Expect(rule.ResourceNames).To(ConsistOf("pull-secret"),
+							"ClusterRole secrets rule must be pinned to exactly pull-secret")
+						Expect(rule.Verbs).To(ConsistOf("get", "list", "watch"),
+							"ClusterRole secrets rule must grant read-only verbs")
+					}
+				}
+			}
+			Expect(found).To(BeTrue(), "ClusterRole must include a pinned pull-secret rule")
+		})
+
+		It("does not grant cluster-wide networkpolicies access", func() {
+			for _, rule := range clusterRole.Rules {
+				for _, res := range rule.Resources {
+					Expect(res).NotTo(Equal("networkpolicies"),
+						"ClusterRole must not grant networkpolicies access — should be in namespaced Role")
+				}
+			}
+		})
+
+		It("does not use create verb with resourceNames for RBAC resources", func() {
+			rbacResources := map[string]bool{
+				"clusterroles":        true,
+				"clusterrolebindings": true,
+				"rolebindings":        true,
+				"roles":               true,
+			}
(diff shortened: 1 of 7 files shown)


internal/rbac/role_yaml_test.go
```diff
@@ -0,0 +1,164 @@
...
+					if res == "secrets" {
+						Expect(rule.ResourceNames).To(ConsistOf("pull-secret"),
+							"ClusterRole secrets rule must be pinned to exactly pull-secret")
+					}
+				}
+			}
+		})
+
+		It("does not grant cluster-wide networkpolicies access", func() {
+			for _, rule := range clusterRole.Rules {
```
blublinsky, line 64: **should-fix**: This test can pass silently without actually verifying anything. If the `pull-secret` rule is accidentally removed from the ClusterRole, the `if res == "secrets"` condition never matches, the loop body never executes, zero assertions fire, and the test reports PASS.

The Role tests in the same file already handle this correctly with a `found` sentinel — the ClusterRole test should do the same:

```go
found := false
for _, rule := range clusterRole.Rules {
    for _, res := range rule.Resources {
        if res == "secrets" {
            found = true
            Expect(rule.ResourceNames).To(ConsistOf("pull-secret"),
                "ClusterRole secrets rule must be pinned to exactly pull-secret")
        }
    }
}
Expect(found).To(BeTrue(), "ClusterRole must include a pinned pull-secret rule")
```

internal/rbac/role_yaml_test.go
```diff
@@ -0,0 +1,164 @@
...
+				for _, res := range rule.Resources {
+					if res == "secrets" {
+						found = true
+						actual := make([]string, len(rule.Verbs))
+						copy(actual, rule.Verbs)
+						sort.Strings(actual)
+						Expect(actual).To(Equal(expected), "Role secrets verbs mismatch")
+					}
+				}
+			}
```
blublinsky, line 177: **should-fix**: The tests validate that `resourceNames` rules don't combine with `create` and that named resources are in an allowlist — but no test verifies the converse: that destructive write verbs (`update`, `delete`) are **always** pinned to `resourceNames`. That's the actual invariant that prevents privilege escalation (e.g. updating an arbitrary ClusterRoleBinding to point to `cluster-admin`).

Suggested addition:

```go
It("requires resourceNames for update/delete verbs on RBAC resources", func() {
    rbacResources := map[string]bool{
        "clusterroles":        true,
        "clusterrolebindings": true,
        "rolebindings":        true,
    }
    restrictedVerbs := map[string]bool{"update": true, "delete": true}

    for _, rule := range clusterRole.Rules {
        hasRBAC := false
        for _, res := range rule.Resources {
            if rbacResources[res] {
                hasRBAC = true
            }
        }
        if !hasRBAC {
            continue
        }
        for _, verb := range rule.Verbs {
            if restrictedVerbs[verb] {
                Expect(rule.ResourceNames).NotTo(BeEmpty(),
                    "RBAC rule with %q verb must be pinned to resourceNames", verb)
            }
        }
    }
})
```

This ensures `create` can stay un-pinned (Kubernetes requirement), but `update`/`delete` must always be scoped to specific named resources.

blublinsky (commented):
**Must-fix: Informer for `openshift-config` secrets will get 403 Forbidden, crashing the operator**

The ClusterRole now only has this rule for secrets:
```yaml
- apiGroups: [""]
  resourceNames: [pull-secret]
  resources: [secrets]
  verbs: [get, list, watch]
```

Since Kubernetes 1.10+ ([kubernetes/kubernetes#63469](https://github.com/kubernetes/kubernetes/pull/63469)), `list` and `watch` verbs in a rule with `resourceNames` require the client to include a matching `metadata.name` field selector. From the [Kubernetes RBAC docs](https://kubernetes.io/docs/reference/access-authn-authz/rbac/):

> If you restrict list or watch by resourceName, clients must include a `metadata.name` field selector in their list or watch request that matches the specified resourceName in order to be authorized.

The operator's cache in `cmd/main.go` uses an empty `cache.Config{}` for `openshift-config` — no field selector:
```go
ByObject: map[client.Object]cache.ByObject{
    &corev1.Secret{}: {
        Namespaces: map[string]cache.Config{
            namespace:                          {},
            utils.TelemetryPullSecretNamespace: {},  // empty config, no field selector
        },
    },
}
```

The informer will issue a bare `LIST /api/v1/namespaces/openshift-config/secrets` (no `?fieldSelector=metadata.name=pull-secret`), which does **not** match the `resourceNames` rule → 403 → informer fails → operator crashes.

Unit tests pass because envtest doesn't enforce RBAC.

**Suggested fixes** (pick one):
1. **Add field selector to cache** (recommended): `TelemetryPullSecretNamespace: {FieldSelector: fields.SelectorFromSet(fields.Set{"metadata.name": "pull-secret"})}` — this makes the LIST request match the RBAC rule precisely.
2. **Keep a separate unrestricted `list;watch` rule**: Add back `resources=secrets, verbs=list;watch` without `resourceNames` at cluster scope (what PR #1944 preserved). Broader but simpler.

PR #1993, opened 2026-08-26
OLS-3548 Expose temperatureSupported model parameter

xrajesh:
## Summary

Exposes the `temperature_supported` model option (added to the service in openshift/lightspeed-service#3000) through the `OLSConfig` CR, so operator-managed clusters can disable the `temperature` parameter for models that reject it (e.g. `claude-sonnet-5`) and avoid HTTP 400 errors.

## Changes

- **API** — new `temperatureSupported *bool` field on `ModelParametersSpec` (`api/v1alpha1/olsconfig_types.go`).
- **Config generation** — `buildProviderConfigs` passes the value through to the generated `olsconfig` `parameters` block (`assets.go`), serialized as `temperature_supported`.
- **Generated artifacts** — regenerated deepcopy; CRD (`config/crd` + `bundle`) and CSV descriptors updated with just the new property.
- **Tests** — `assets_test.go` asserts the key is omitted when unset (service default applies) and propagated as `false` when set.

## Design

`*bool` with `omitempty` is deliberate. The service default is `true`, so a plain `bool` could not distinguish *unset* from an explicit `false`. With a pointer:

- **unset** → key omitted from the generated config → the service applies its own default (`true`). Single source of truth for the default; no risk of the operator's default drifting from the service.
- **explicit `true`/`false`** → propagated verbatim.

This matches the existing `*bool` precedent in the CRD (`logging`, `introspectionEnabled`, `auditEventsEnabled`).

## Usage

```yaml
spec:
  llm:
    providers:
      - name: my-bedrock
        models:
          - name: anthropic.claude-sonnet-5
            parameters:
              temperatureSupported: false
```

## Test plan

- [x] `make test` passes (appserver package includes new coverage)
- [x] `make generate manifests` produces the deepcopy/CRD changes (unrelated pre-existing TLS-profile regeneration drift excluded to keep the diff focused)

## Note

The operator still mirrors only a subset of the service's `ModelParameters` (`max_tokens_for_response`, `tool_budget_ratio`, and now `temperature_supported`); `reasoning_effort`, `reasoning_summary`, `verbosity`, and `temperature` remain unexposed. A generic passthrough to close that gap systematically is worth a separate RFE.

## Summary by CodeRabbit

* **New Features**
  * Added configuration support to indicate whether a model accepts temperature parameters.
  * Exposed the “Temperature Supported” setting in the configuration interface.
  * Preserved explicit `false` values while omitting the setting when unspecified.
  * Applied the setting to generated application-server configuration for consistent model behavior.
* **Documentation**
  * Updated configuration descriptors to explain the setting and its default behavior.
  * Updated the operator package version to 1.1.4.


xrajesh:
LGTM

PR #2013, opened 2026-09-03
OLS-2190: require username and version for on-prem watsonx secrets

DeokarT:
[OLS-2190](https://redhat.atlassian.net/browse/OLS-2190). Also OLS-2849.

IBM Cloud watsonx stays as it is: secret key `apitoken`. Empty URL is treated as IBM Cloud because the service default is `https://us-south.ml.cloud.ibm.com`.

If the provider URL is not `*.ml.cloud.ibm.com`, this is Cloud Pak for Data. Then the same secret also needs `username` and `version`. `instance_id` is optional. The operator already mounts the whole secret, so those keys are already on disk. We just start requiring them instead of letting the app die on `WATSONX_USERNAME` and then on version.

Do not tell people to set env vars on the app Deployment. We will revert that.

Service PR: https://github.com/openshift/lightspeed-service/pull/3081

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,12 +36,28 @@ IMG="docker.io/username/ols-operator:0.10" make deploy
 2. Create a secret containing the API Key for Watsonx, OpenAI, Azure OpenAI. The key for API key is `apitoken`.
 
 > [!TIP]
-> Watsonx example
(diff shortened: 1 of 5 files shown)


onmete:
WantsonX on-prem is deferred - the team has no capacity to proceed here.
/close

PR #2014, opened 2026-09-03
OLS-3634: ask command with SSE streaming

xiormeesh:
## Description

Core user-facing feature for oc-ols: send a question to OLS and stream back the answer via SSE.

```bash
oc ols "why is my pod crashing?"
oc ols ask "how do I create a route?"
```

**Changes:**

- `cli/ask.go` — `AskOptions` with Complete/Validate/Run, token extraction from JSON payloads, referenced documents on stdout
- `cli/streaming.go` — `SSEClient` with HTTP transport, SSE parser for lightspeed-service JSON envelope format (`{"event": "...", "data": ...}`), idle timeout (120s), HTTP status → user-facing error mapping (401/403/413)
- `cli/types.go` — `LLMRequest`, `StartEventData`, `TokenEventData`, `EndEventData`, `sseEnvelope` with `json.RawMessage`
- `cli/root.go` — default mode dispatch (unrecognized args → ask), ask registered as subcommand
- `cli/kubeconfig.go` — improved error for missing kubeconfig using `clientcmd.IsEmptyConfig()`
- `cli/integration_test.go` — end-to-end test through full cobra dispatch
- 63 Ginkgo specs across unit and integration tests

## Type of change

- [x] New feature

- [OLS-3634](https://redhat.atlassian.net/browse/OLS-3634)

## Testing

**Unit/integration tests:** 63 Ginkgo specs, 82% statement coverage on `cli/` package.

**Manual verification against live OLS cluster:**

1. Build: `go build -o bin/oc-ols ./cmd/oc-ols/`
2. Login: `oc login`
3. Set endpoint: `./bin/oc-ols config set-endpoint https://<ols-route-host>`
4. Ask: `./bin/oc-ols "what can OLS do for me?"`

Verified:
- default mode (`oc ols "question"`) and explicit ask (`oc ols ask "question"`)
- token streaming to stdout, referenced documents after response
- HTTP error mapping: 401 → auth error with `Try: oc login`, 403, 413, unreachable endpoint
- empty query rejection, no-endpoint-configured error
- Ctrl+C cancellation during streaming
- pipe to file/other commands (non-TTY behavior)
- TLS flags (`--insecure-skip-tls-verify`, `--ca-cert`)
- version and config subcommands still work
- compared CLI vs UI backend logs — identical server-side processing

## Summary by CodeRabbit

- **New Features**
  - Added the `ask` command for submitting questions to OpenShift Lightspeed.
  - Responses stream to the terminal as they are generated.
  - Referenced documents appear with completed responses.
  - Supports kubeconfig authentication and secure service connections.
  - Allows explicitly permitted HTTP endpoints for development or testing.

- **Bug Fixes**
  - Added clear handling for authentication, access, timeout, malformed-stream, and service errors.
  - Empty questions and missing configuration now provide actionable guidance.
  - Improved handling of incomplete responses and large streamed payloads.

### DIFF
--- a/cli/ask.go
+++ b/cli/ask.go
@@ -1,0 +1,189 @@
+package cli
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+)
+
+const (
+	ErrQueryEmpty       = "no query provided"
+	ErrNoEndpoint       = "endpoint not resolved"
+	ErrStreamIncomplete = "response may be incomplete (stream interrupted)"
+	ErrMalformedEnd     = "failed to parse end event"
+	ErrMissingEnd       = "stream ended without end event"
+)
+
+// AskOptions holds the configuration for the ask command.
+type AskOptions struct {
+	streams    genericclioptions.IOStreams
+	query      string
+	endpoint   string
+	kubeConfig *KubeConfig
+	mode       string
+}
+
+// NewAskCmd creates the "ask" subcommand that sends a question to OLS
+// and streams back the response.
+func NewAskCmd(streams genericclioptions.IOStreams) *cobra.Command {
+	o := &AskOptions{
+		streams: streams,
+		mode:    "ask",
+	}
+
+	cmd := &cobra.Command{
+		Use:   "ask [question]",
+		Short: "Ask OpenShift Lightspeed a question",
+		Long:  "Send a question to OpenShift Lightspeed and stream the response.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := o.Complete(cmd, args); err != nil {
+				return err
+			}
+			if err := o.Validate(); err != nil {
+				return err
+			}
+			return o.Run(cmd)
+		},
+		Args:         cobra.ArbitraryArgs,
+		SilenceUsage: true,
+	}
+
+	return cmd
+}
+
+// Complete resolves the query string, kubeconfig, and endpoint.
+func (o *AskOptions) Complete(cmd *cobra.Command, args []string) error {
+	o.query = strings.Join(args, " ")
+
+	kubeconfigPath, _ := cmd.Flags().GetString("kubeconfig")
+	contextName, _ := cmd.Flags().GetString("context")
+	insecureSkipTLS, _ := cmd.Flags().GetBool("insecure-skip-tls-verify")
+	caCertPath, _ := cmd.Flags().GetString("ca-cert")
+
+	kc, err := LoadKubeConfig(kubeconfigPath, contextName, insecureSkipTLS, caCertPath)
+	if err != nil {
+		return err
+	}
+	o.kubeConfig = kc
+
+	endpoint, err := ResolveEndpoint(cmd, kc.ContextName)
+	if err != nil {
+		return err
+	}
+	o.endpoint = endpoint
+
+	return nil
+}
+
+// Validate checks that required fields are populated.
+func (o *AskOptions) Validate() error {
+	if strings.TrimSpace(o.query) == "" {
+		return fmt.Errorf("%s: provide a question as arguments", ErrQueryEmpty)
+	}
+	if o.endpoint == "" {
+		return errors.New(ErrNoEndpoint)
+	}
+	return nil
+}
+
+// Run executes the ask command: sends the query via SSE and streams
+// token data to stdout. Referenced documents from the end event are
+// printed to stdout along with the response.
+func (o *AskOptions) Run(cmd *cobra.Command) error {
+	client := NewSSEClient(o.endpoint, o.kubeConfig.BearerToken, o.kubeConfig.TLSConfig)
+
+	req := LLMRequest{
+		Query:     o.query,
+		Mode:      o.mode,
+		MediaType: "application/json",
+	}
+
+	ctx := cmd.Context()
+	if ctx == nil {
+		ctx = context.Background()
+	}
+
+	events, errc, err := client.StreamQuery(ctx, req)
+	if err != nil {
+		return err
+	}
+
+	var endData *EndEventData
+	var hasTokens bool
+	var endParseErr error
+
+	for ev := range events {
+		switch ev.Type {
+		case EventToken:
+			var td TokenEventData
+			if err := json.Unmarshal([]byte(ev.Data), &td); err != nil {
+				// If token data isn't JSON, use raw string as fallback
+				td.Token = ev.Data
+			}
+			hasTokens = true
+			if _, err := fmt.Fprint(o.streams.Out, td.Token); err != nil {
+				return fmt.Errorf("%s: %w", ErrWriteOutput, err)
+			}
+		case EventEnd:
+			var ed EndEventData
+			if err := json.Unmarshal([]byte(ev.Data), &ed); err != nil {
+				endParseErr = err
+			} else {
+				endData = &ed
+			}
+		// start: conversation_id bookkeeping (used in PR5/OLS-3636)
+		// reasoning, tool_call, tool_result: not displayed in default mode.
+		// TODO(OLS-3639): accumulate for --output json
+		default:
+		}
+	}
+
+	// Check for stream-level errors
+	if streamErr := <-errc; streamErr != nil {
+		if _, err := fmt.Fprintf(o.streams.ErrOut, "Warning: %s\n", ErrStreamIncomplete); err != nil {
+			return fmt.Errorf("%s: %w", ErrWriteOutput, err)
+		}
+		return streamErr
+	}
+
+	// Print trailing newline only if tokens were emitted
+	if hasTokens {
+		if _, err := fmt.Fprintln(o.streams.Out); err != nil {
+			return fmt.Errorf("%s: %w", ErrWriteOutput, err)
+		}
+	}
+
+	// Malformed or missing end event means the stream was not fully valid
+	if endParseErr != nil {
+		if _, err := fmt.Fprintf(o.streams.ErrOut, "Warning: %s: %v\n", ErrMalformedEnd, endParseErr); err != nil {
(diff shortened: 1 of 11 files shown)


cli/streaming.go
```diff
@@ -0,0 +1,291 @@
...
+		return nil, nil, fmt.Errorf("%s: %w", ErrSendRequest, err)
+	}
+
+	httpReq, err := http.NewRequestWithContext(ctx, http.MethodPost,
+		c.endpoint+streamingQueryPath, bytes.NewReader(body))
+	if err != nil {
+		return nil, nil, fmt.Errorf("%s: %w", ErrSendRequest, err)
+	}
+	httpReq.Header.Set("Content-Type", "application/json")
+	httpReq.Header.Set("Authorization", "Bearer "+c.token)
```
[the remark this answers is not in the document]
xiormeesh: Addressed in 3964fda — Validate() now rejects cleartext HTTP endpoints unless --insecure-skip-tls-verify is passed. This covers both --endpoint flag and persisted values.

onmete:
`parseSSEStream` ignores `event:` fields and always treats `data:` as the custom JSON envelope. That does not satisfy the SSE event/data-line acceptance criterion: an `event: token` + `data: {...}` frame will fail envelope parsing. Please parse the SSE event field and payload, or explicitly support both that framing and the current OLS JSON-envelope framing.

onmete:
`Run` discards reasoning, tool_call, and tool_result frames in the default branch. [OLS-3634](https://redhat.atlassian.net/browse/OLS-3634) requires these events to be captured while remaining hidden from default output. Please retain their payloads in command/result state and keep stdout limited to the response tokens and references.

onmete:
`Run` never retains the conversation ID: `StartEventData` is defined but start frames are discarded, and `EndEventData` has no conversation-ID field. [OLS-3634](https://redhat.atlassian.net/browse/OLS-3634) requires extracting the returned conversation ID. Please parse and retain it from the service’s start event so it is available for conversation handling.

onmete:
Those are three ACs that are not met from the linked [OLS-3634](https://redhat.atlassian.net/browse/OLS-3634).

cli/ask.go
```diff
@@ -0,0 +1,202 @@
...
+		return err
+	}
+	o.kubeConfig = kc
+	o.insecureAllowHTTP = insecureSkipTLS
+
+	endpoint, err := ResolveEndpoint(cmd, kc.ContextName)
+	if err != nil {
+		return err
+	}
+	o.endpoint = endpoint
```
raptorsun, line 91: `--insecure-skip-tls-verify` is being overloaded here to also allow cleartext HTTP endpoints. However, `config set-endpoint` already uses a dedicated `--insecure-allow-http` flag for this purpose (`cli/config/endpoint.go:91`), per the spec's distinction:

- `--insecure-skip-tls-verify`: skip TLS cert verification on HTTPS (e.g., self-signed certs)
- `--insecure-allow-http`: allow sending bearer token over cleartext HTTP (dev/test only)

These are different security decisions — a user skipping TLS verification for a self-signed HTTPS endpoint doesn't necessarily want to send their token unencrypted over HTTP.

Suggestion: add `--insecure-allow-http` as a flag on the ask command (or promote it to a global persistent flag) and use it here instead, so the ask and config commands are consistent.

cli/streaming.go
```diff
@@ -0,0 +1,291 @@
...
+
```
raptorsun, line 114: Nit: consider adding `Accept: text/event-stream` here. While the `/v1/streaming_query` endpoint returns SSE regardless, setting the Accept header is good HTTP practice — it makes the client's intent explicit and ensures correct behavior if a proxy or future server version uses content negotiation.

```suggestion
	httpReq.Header.Set("Content-Type", "application/json")
	httpReq.Header.Set("Accept", "text/event-stream")
	httpReq.Header.Set("Authorization", "Bearer "+c.token)
```

raptorsun:
> `Run` never retains the conversation ID...

Conversation ID retention is explicitly scoped to [OLS-3636](https://redhat.atlassian.net/browse/OLS-3636), not this PR. The spec states: *"Conversation ID persistence planned for [OLS-3636](https://redhat.atlassian.net/browse/OLS-3636)."* (`cli.md`, Conversation persistence section)

The infrastructure is already in place for [OLS-3636](https://redhat.atlassian.net/browse/OLS-3636) to build on:
- `StartEventData.ConversationID` is defined in `cli/types.go`
- `LLMRequest.ConversationID` is defined with `omitempty` for the request path

Also note that in the actual lightspeed-service wire format, `conversation_id` comes in the **start** event, not the end event. The AC's phrasing "End event extracts conversation_id" is slightly inaccurate — `EndEventData` correctly models what the end event actually contains (referenced_documents, token counts, truncated flag).

cli/ask.go
luka-j2: Confirmed and fixed in 754db42. `publish_service.py:938-940` is a bare `self.standard.save()` on an instance loaded at the start of the job, and publish populates BigQuery tables before it gets there — so the window is minutes, and the rename loses silently after the table has already said "Standard renamed".

This is the same invariant the row menu already encodes twelve lines below, which is what makes it worth taking rather than arguing: *"Neither is offered mid-publish — the running job holds a `Standard` loaded before the delete and saves it back, so the row would silently return minutes later."* Rename was the one write on the row that had not picked the rule up. The name cell now renders the plain name for a `processing` row, matching the absent row menu rather than inventing a disabled affordance next to it.

Grepped for the rest of the class while I was in there: `useStandardArchiveMutations` covers archive, delete and restore, and the row menu already withholds all three mid-publish. Rename was the only gap. `error` rows stay editable — there is no job running against one.

Server-side concurrency safety is the other half of your suggestion and I have not done it: making publish tolerate a concurrent write means either a narrower `save(update_fields=...)` or re-reading under a lock, in a service that four other flows call. That belongs with APP-5781, which is already adding the server-side guards around editing a standard that is not a draft.

luka-j2: Confirmed and fixed in 754db42. `publish_service.py:938-940` is a bare `self.standard.save()` on an instance loaded at the start of the job, and publish populates BigQuery tables before it gets there — so the window is minutes, and the rename loses silently after the table has already said "Standard renamed".

This is the same invariant the row menu already encodes twelve lines below, which is what makes it worth taking rather than arguing: *"Neither is offered mid-publish — the running job holds a `Standard` loaded before the delete and saves it back, so the row would silently return minutes later."* Rename was the one write on the row that had not picked the rule up. The name cell now renders the plain name for a `processing` row, matching the absent row menu rather than inventing a disabled affordance next to it.

Grepped for the rest of the class while I was in there: `useStandardArchiveMutations` covers archive, delete and restore, and the row menu already withholds all three mid-publish. Rename was the only gap. `error` rows stay editable — there is no job running against one.

Server-side concurrency safety is the other half of your suggestion and I have not done it: making publish tolerate a concurrent write means either a narrower `save(update_fields=...)` or re-reading under a lock, in a service that four other flows call. That belongs with APP-5781, which is already adding the server-side guards around editing a standard that is not a draft.

xiormeesh: huh not sure how it got mixed up, fixed now in `9258ade`, `--insecure-allow-http` is now a separate global persistent flag, consistent with config set-endpoint. `--insecure-skip-tls-verify` only controls TLS cert verification. thank you for spotting it!

cli/streaming.go
xiormeesh: good defensive approach, done in `9258ade`

xiormeesh:
@onmete `9258ade` your 3 finds, thanks for flagging it! 
**SSE event: field parsing**
`parseSSEStream` now parses `event:` fields. When present, uses it as the event type and passes data: through as-is. When absent, falls back to JSON envelope extraction. Both framings are tested.
**reasoning/tool events discarded**
Reasoning, tool_call, and tool_result events are accumulated in `capturedEvents` instead of discarded. Not displayed in default mode, available for `--output json` in OLS-3639.
**conversation_id not retained**
Start events are parsed and `conversation_id` is extracted into `AskOptions.conversationID`. Persistence is OLS-3636, this PR handles extraction. (As Haoyu noted in his earlier reply)

cli/ask.go
```diff
@@ -0,0 +1,222 @@
...
+			if _, err := fmt.Fprint(o.streams.Out, td.Token); err != nil {
+				return fmt.Errorf("%s: %w", ErrWriteOutput, err)
+			}
+		case EventStart:
+			// Extract conversation_id for persistence (OLS-3636).
+			var sd StartEventData
+			if err := json.Unmarshal([]byte(ev.Data), &sd); err == nil {
+				o.conversationID = sd.ConversationID
+			}
+			o.capturedEvents = append(o.capturedEvents, ev)
```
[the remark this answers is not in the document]
xiormeesh: This is a CLI that runs once per invocation. The stream ends naturally (end event) and has a 120s idle timeout. Even agentic responses with multiple tool rounds produce ~10-20 events. OOM from captured events is not a realistic scenario for a short-lived CLI process.

raptorsun:
> `parseSSEStream` ignores `event:` fields and always treats `data:` as the custom JSON envelope. That does not satisfy the SSE event/data-line acceptance criterion...

The lightspeed-service does **not** send standard SSE `event:`/`data:` field framing. It sends a custom JSON envelope inside the `data:` line:

```
data: {"event": "token", "data": {"id": 0, "token": "Hello"}}

```

The event type is embedded inside the JSON payload, not as a separate SSE `event:` field. The parser (`parseSSEEnvelope`) correctly extracts the event type and inner data from this envelope format. The spec describes this explicitly (`cli.md`, `parseSSEStream` docstring):

> *lightspeed-service sends events as JSON envelopes inside SSE data lines: `{"event": "<type>", "data": <payload>}`*

The spec line "Parse SSE event frames (`event:`, `data:` lines)" is describing SSE generically — the parenthetical lists SSE concepts, not the literal wire format. The PR author verified this against a live OLS cluster, and the test data matches the actual server output format.

An `event: token` + `data: {...}` frame (standard SSE framing) is not what the server sends, so there is no need to parse it.

PR #2018, opened 2026-09-04
OLS-3548 Regenerate bundle for temperatureSupported

xrajesh:
Depends on #1993 and should be merged after it.

This PR contains only the generated bundle manifest updates for temperatureSupported, split out so #1993 can run/pass the operator tests first.

## Summary by CodeRabbit

* **Documentation**
  * Documented the model temperature support setting.
  * Clarified that leaving the setting unset defaults to supported, while disabling it applies to models that reject temperature parameters.


xrajesh:
@sriroopar @JoaoFula I am overriding this , but we will have to address the bugs identified.

xrajesh:
/override "Red Hat Konflux / service-e2e-tests-419 / ols-bundle"

PR #2028, opened 2026-09-07
OLS-3453 OLS-3461: Add reasoningConfig CRD field and release notes

blublinsky:
### Summary
  
  Add provider-agnostic `reasoningConfig` field to `ModelParametersSpec` for configuring reasoning/thinking parameters per model without code changes. Field is passed through to service for
  provider-specific validation. Includes comprehensive release notes with provider reference table and examples.
  
  ### Changes
  
  - Add `ReasoningConfig map[string]interface{}` to CRD `ModelParametersSpec`
  - Mirror `ReasoningConfig` field in internal `ModelParameters` type
  - Wire `ReasoningConfig` through config generation pipeline
  - Include `reasoning_config` in generated `olsconfig.yaml`
  - Update test to avoid comparing non-comparable map types
  - Regenerate bundle with updated CRD and CSV manifests
  - Add release notes documenting valid keys per provider and examples
  
  ### Supported Providers
  
  - **OpenAI** (gpt-5-x): `effort`, `summary`, `verbosity`
  - **Anthropic Claude**: `thinking_enabled`, `budget_tokens`
  - **Google Gemini**: `thinking_level`, `thinking_budget_tokens`

## Type of change

- [ ] Refactor
- [ x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing

## Summary by CodeRabbit

- **New Features**
  - Added optional reasoning configuration for model parameters.
  - Reasoning settings are provider-agnostic and support provider-specific options validated by the model service.
  - Configuration is exposed in the operator settings and passed through to the configured model service.

- **Documentation**
  - Updated configuration references and operator descriptors to document reasoning configuration.
  - Clarified deployment settings and replica configuration descriptions in the operator interface.


config/crd/bases/ols.openshift.io_olsconfigs.yaml
```diff
@@ -289,6 +289,11 @@ spec:
                                     description: Max tokens for response. The default
                                       is 2048 tokens.
                                     type: integer
+                                  reasoningConfig:
+                                    description: |-
+                                      Reasoning configuration for the model (provider-agnostic freeform config).
+                                      The service and provider API validate the contents.
+                                    type: object
```
vimalk78, line 296: This freeform map needs to preserve arbitrary provider-specific keys through the Kubernetes API. The generated CRD only has `type: object` and does not include `x-kubernetes-preserve-unknown-fields: true` (or an equivalent schema). Without preservation, keys such as `effort`, `thinking_budget`, and `include_thoughts` may be pruned before the operator generates `olsconfig.yaml`. Please use the correct kubebuilder marker and regenerate the CRD, then add coverage for a nested reasoningConfig value.

config/manifests/bases/lightspeed-operator.clusterserviceversion.yaml
```diff
@@ -338,9 +344,18 @@ spec:
...
         - urn:alm:descriptor:com.tectonic.ui:podCount
-      - description: RHOKP sidecar container settings (Solr / OKP).
+      - description: RHOKP standalone deployment settings (Solr / OKP).
         displayName: RHOKP Container
         path: ols.deployment.rhokp
+      - description: |-
+          Defines the number of desired pods. Default: "1"
+          Note: Replicas are configurable for APIContainer and MCP server (mcpServer).
+          For PostgreSQL, Console, Agentic Console, Alerts Adapter, OTEL Collector, and
+          Agentic Sandbox (spec.agenticOLS.agenticSandboxConfig), the number of replicas is always set to 1.
```
vimalk78, line 354: The new RHOKP replica descriptor says which components are always set to one replica, but omits RHOKP itself. Please include RHOKP in this list so the UI metadata does not imply that `ols.deployment.rhokp.replicas` is configurable.

vimalk78 (changes requested):
## Review summary

I found two requirement issues and one documentation issue.

- The generated CRD does not preserve arbitrary `reasoningConfig` keys.
- OLS-3461 requires user-facing release notes and documentation with provider tables, examples, migration guidance, and behavior details, but those documents are not included in this diff.
- The RHOKP fixed-replica descriptor omits RHOKP from its own always-one-replica description.

The detailed review is saved locally in `.ols/reviews/pr-2028.md`. Existing PR CI reports all tests passed.

config/crd/bases/ols.openshift.io_olsconfigs.yaml
blublinsky: fixed

config/manifests/bases/lightspeed-operator.clusterserviceversion.yaml
blublinsky: This is a generated file. This PR does not touch RHOKP

vimalk78 (commented):
## Findings

### [P1] Do not downgrade the bundle version to 1.1.2

`bundle.Dockerfile:36` and `bundle.Dockerfile:39` change the bundle labels from `1.1.4` to `1.1.2`, while the generated CSV still identifies the operator as `1.1.3` (`bundle/manifests/lightspeed-operator.clusterserviceversion.yaml:58` and `:1177`). This creates inconsistent bundle metadata and advertises a downgrade. Please regenerate the bundle with the intended version and ensure the Dockerfile labels and CSV metadata match.

### [P2] Add deepcopy handling for `ReasoningConfig`

The new `map[string]interface{}` field at `api/v1alpha1/olsconfig_types.go:551` is not copied in `api/v1alpha1/zz_generated.deepcopy.go:297-304`. `DeepCopyInto` currently performs only `*out = *in`, so the map and nested maps remain aliased when callers use `OLSConfig.DeepCopy()`. Mutating a copied object can mutate the original cached object or test fixture. The added test only verifies API round-tripping, not deepcopy isolation. Please add recursive deepcopy handling (or use a type for which controller-gen generates it) and test that mutating the copy does not change the original.

`make generate` did not add deepcopy handling. The full `make test` run was blocked while downloading external CRDs from `raw.githubusercontent.com`.

vimalk78 (commented):
## Follow-up findings

### [P2] Deep-copy nested `ReasoningConfig` values

The new `DeepCopyInto` implementation copies the top-level map but assigns each interface value directly (`api/v1alpha1/zz_generated.deepcopy.go:306-308`). This still aliases nested maps, slices, or other reference values. The existing reconciliation test uses a nested map, but the new deepcopy unit test only uses scalar values, so it would not catch this. For example, after `out := in.DeepCopy()`, mutating `out.ReasoningConfig["nested"].(map[string]interface{})["key"]` also mutates `in`. Please recursively copy arbitrary JSON values and add a nested-map/slice assertion.

### [P2] Document RHOKP as fixed at one replica

The new OLM descriptor for `ols.deployment.rhokp.replicas` in `config/manifests/bases/lightspeed-operator.clusterserviceversion.yaml` says replicas are configurable for APIContainer and MCP server, then lists PostgreSQL, Console, Agentic Console, Alerts Adapter, OTEL Collector, and Agentic Sandbox as fixed at one. RHOKP is missing from that fixed-replica list, even though the operator forces RHOKP replicas to one. This can mislead users about a setting the operator does not honor. Add RHOKP to the fixed-replica list and regenerate the bundle.

